PR #83, opened 2022-08-31
TEP: FunC project configuration file

dvlkv:


tvorogme:
Discussed with @dvlkv in private conversation about adding aliases to each entry point. So any third-party tools can add own modules to project file.

text/0083-func-config-standard.md
```diff
@@ -0,0 +1,66 @@
...
+Also, here comes more convinient project bootstapping. 
+
+# Specification
+
+I suppose it should be yaml file, named like .func.yml or something else.
+In first implementation it should contain:
+- root source folder
+- entrypoints (contract root files), relative to source folder
+- (optional) library directories
+- (optional) compiler configuration (version, etc.)
```
tvorogme, line 28: I think it's must be not optional, just need to add version of compiler for any. While each version can present new syntax.

text/0083-func-config-standard.md
```diff
@@ -0,0 +1,66 @@
...
+- (optional) library directories
+- (optional) compiler configuration (version, etc.)  
+
+Example file looks like:
+```yml
+сompiler: 
+    version: 0.2.0
+sources: src/
+entrypoints:
+    - contracts/nft.fc
```
tvorogme, line 36: Need to add aliases as @dvlkv suggested.

text/0083-func-config-standard.md
```diff
@@ -0,0 +1,66 @@
...
+    - contracts/nft-collection.fc
+libraries:
+    - node_modules/func-*
+```
```
tvorogme, line 40: Need example of aliases usage.

text/0083-func-config-standard.md
```diff
@@ -0,0 +1,66 @@
...
+
+Extensions can use it for optimal indexing source code, compilers can be configured without bash scripts.  
+Also, here comes more convinient project bootstapping. 
+
+# Specification
+
+I suppose it should be yaml file, named like .func.yml or something else.
+In first implementation it should contain:
+- root source folder
+- entrypoints (contract root files), relative to source folder
```
tvorogme, line 26: I suggest to add `(optional) metadata` so it'll be easy to publish repos somewhere

```
meta:
    name: "My cool project"
    version: "0.0.1"
    description: "NFT to the moon"
    repository: "uri://my_cool_project"
    homepage: "uri://my_cool_project/docs"
    license: "WTFPL"
    author: "Cool Guy"

    contracts:
      collection:
        name: "Collection base smart contract"
        description: "Just cool collection smart"

      nft:
        name: "Nft base smart contract"
        description: "It's used in collection smart contract in c4, so you need to pass this code to collection"
```

text/0083-func-config-standard.md
```diff
@@ -0,0 +1,66 @@
...
+Extensions, compilers and other dev tools need universal project configuration.
+
+# Guide
+
+Extensions can use it for optimal indexing source code, compilers can be configured without bash scripts.  
+Also, here comes more convinient project bootstapping. 
+
+# Specification
+
+I suppose it should be yaml file, named like .func.yml or something else.
```
tvorogme, line 23: I think we need one unique name `ton.yml` without any other options. `func.yml` also ok, but I prefer `ton.yml`, it's really cool :)

Since most file systems hide files started with `.` I suggest not to use it.

text/0083-func-config-standard.md
```diff
@@ -0,0 +1,66 @@
...
+
+# Drawbacks
+
+Existing tools with configuration files should be adopted to work correct with this format.
+
+# Rationale and alternatives
+
+- Why is this design the best in the space of possible designs?
+
+In this implementation, configuration file contains only generic information without domain-specific data.
```
tvorogme, line 50: I think the main idea here is that this structure can be modular and other new third-party tools can simply add own module.

tvorogme (changes requested):

text/0083-func-config-standard.md
EmelyanenkoK: I believe that `func` in name is not optimal, since it will be better if one config format will be able to handle projects on different languages: func, tact, fift-asm, etc. 
`ton` is strange as well, why not `project`?

text/0083-func-config-standard.md
dvlkv: Compiler version can be declared using #pragma, so I suppose that it is just an alternative to it.

text/0083-func-config-standard.md
dvlkv: Would be better to have an additional standard, there are dozens of things to discuss only about meta.

text/0083-func-config-standard.md
dvlkv: We can collect more variants and make a poll ¯\_(ツ)_/¯

text/0083-func-config-standard.md
tvorogme: got it. But my idea is to force all devs to write down version of compiler. It can be that you found project and can't compile it because it miss-versioned. I have encountered this very often. This little action will make life very easy.

text/0083-func-config-standard.md
tvorogme: You think so? It seems to me that it breeds entities out of thin air. It feels like a small number of fields is enough.

text/0083-func-config-standard.md
tvorogme: This is more @andreypfau idea about naming config like `project` confusing while you have 100500 projects on different languages, so the name must be more ton-related. 

I got idea about diffrent languages, Didn't think of that at first.

hacker-volodya: `contract` or `contracts` maybe?

text/0083-func-config-standard.md
tvorogme: Another possible sense idea would be to describe all the extensions in this md file :)

text/0083-func-config-standard.md
andreypfau: ton-project.yml?

EmelyanenkoK: actor? I believe it will be better to migrate away from contract term in the long-term future.

dvlkv:
@EmelyanenkoK can we create a poll here about filename? One thing is about dot prefix, and second one with variants: func.yml, ton.yml, project.yml, tonproj.yml, ton-project.yml

MrG9090 (approved):

Azadm000:
> _ No description provided. _

Azadm000:
https://github.com/ton-blockchain/TEPs/pull/83#issue-1357640925

PR #85, opened 2022-09-01
TEP: SBT Standard

xssnick:
Soul bound token (SBT) is a special kind of NFT which can be transferred only between its owner's accounts. For this, it stores immutable public key of the owner, and it is needed to send transfer from new address with signature in payload to change owner's address.

There is a useful type of token which allows to give social permissions/roles or certificates to some users. For example, it can be used by marketplaces to give discounts to owners of SBT, or by universities to give attestation certificates in SBT form. Mechanics with ownership proof allows to easily prove to any contract that you are an owner of some SBT.

SBT implements [NFT standard interface](https://github.com/ton-blockchain/TIPs/issues/62) but `transfer` should always be rejected, `pull_ownership` is used instead.

### DIFF
--- a/text/0083-sbt-standard.md
+++ b/text/0083-sbt-standard.md
@@ -1,0 +1,255 @@
+- **TEP**: [0](https://github.com/ton-blockchain/TEPs/pull/0)
+- **title**: SBT Contract
+- **status**: Draft
+- **type**: Contract Interface
+- **authors**: [Oleg Baranov](https://github.com/xssnick), [Narek Abovyan](https://github.com/Naltox), [Kirill Emelyanenko](https://github.com/EmelyanenkoK)
+- **created**: 09.08.2022
+
+# Summary
+
+Soul bound token (SBT) is a special kind of NFT which can be transferred only between its owner's accounts. For this, it stores immutable public key of the owner, and it is needed to send transfer from new address with signature in payload to change owner's address.
+
+# Motivation
+
+There is a useful type of token which allows to give social permissions/roles or certificates to some users. For example, it can be used by marketplaces to give discounts to owners of SBT, or by universities to give attestation certificates in SBT form. Mechanics with ownership proof allows to easily prove to any contract that you are an owner of some SBT.
+
+# Specification
+
+SBT implements [NFT standard interface](https://github.com/ton-blockchain/TIPs/issues/62) but `transfer` should always be rejected, `pull_ownership` is used instead.
+
+#### 1. `pull_ownership`
+
+TL-B schema of inbound message:
+```
+pull_ownership#03fdd6c9 query_id:uint64 signature:^(signature:(bits 512)) 
+sbt_nonce:uint64 new_owner:MsgAddress response_destination:MsgAddress 
+custom_payload:(Maybe ^Cell)
+```
+`query_id` -  arbitrary request number.
+
+`signature` - signature of the rest part of the message.
+
+`sbt_nonce` - nonce, required for protection of signature replay attacks.
+
+`new_owner` - address of the new owner of the SBT item, should be the same as sender's address.
+
+`response_destination` - address where a response with confirmation of a successful pull and the rest of the incoming message coins should be sent.
+
+`custom_payload` - optional custom data.
+
+**Should be rejected if:**
+1. `signature` verification failed.
+2. `sbt_nonce` not equals stored nonce.
+3. `new_owner` is not sender's address.
+4. there is not enough coins (with respect to NFT own storage fee guidelines).
+5. After processing the request, the contract must send at least in_msg_value - forward_amount - max_tx_gas_price to the response_destination address. If the contract cannot guarantee this, it must immediately stop executing the request and throw the error.
+
+**Otherwise should do:**
+1. Change current owner of SBT to `new_owner` address.
+2. Generate and store new random `nonce`.
+3. Send all excesses of incoming message coins to response_destination with the following layout:
+   
+   `excesses#d53276db query_id:uint64 = InternalMsgBody;`
+
+   query_id should be equal with request's `query_id`.
+
+#### 2. `prove_ownership`
+
+TL-B schema of inbound message:
+```
+prove_ownership#38061b82 query_id:uint64 dest:MsgAddress 
+data:^Cell with_content:bool
+```
+`query_id` -  arbitrary request number.
+
+`dest` -  address of the contract to which the ownership of SBT should be proven.
+
+`data` - arbitrary data required by target contract.
+
+`with_content` - if true, SBT's content cell will be included in message to contract.
+
+**Should be rejected if:**
+Sender address is not an owner address.
+
+**Otherwise should do:**
+
+Send message with TL-B schema to `dest` contract:
+```
+verify_ownership#01b628aa query_id:uint64 sbt_id:uint256 owner:MsgAddress 
+data:^Cell content:(Maybe ^Cell)
+```
+
+`query_id` - request number passed in `prove_ownership`.
+
+`sbt_id` -  id of SBT.
+
+`owner` - current owner's address.
+
+`data` - data cell passed in `prove_ownership`.
+
+`content` - SBT's content, it is passed if `with_content` was true in `prove_ownership`.
+
+In case when `verify_ownership` was bounced back to SBT, SBT should send message to owner with schema:
+```
+verify_ownership_bounced#450cc71b query_id:uint64 sbt_id:uint256 owner:MsgAddress 
+data:^Cell content:(Maybe ^Cell)
+```
+
+#### 3. `destroy`
+
+TL-B schema of inbound message:
+```
+destroy#2124b0b9 query_id:uint64
+```
+`query_id` -  arbitrary request number.
+
+**Should be rejected if:**
+Sender address is not an owner's address.
+
+**Otherwise should do:**
+Set owner's address to null and set public key to 0.
+
+#### 4. `revoke`
+
+TL-B schema of inbound message:
+```
+revoke#6a6f099e query_id:uint64
+```
+`query_id` -  arbitrary request number.
+
+**Should be rejected if:**
+Sender address is not an authority's address.
+
+**Otherwise should do:**
+Set owner's address to null and set public key to 0.
+
+**GET methods**
+1. `get_public_key()` - returns `int`, that is owner's public key.
+2. `get_nonce()` - returns `int`, which current nonce.
+3. `get_nft_data()` - same as in [NFT standard](https://github.com/ton-blockchain/TIPs/issues/62).
+4. `get_authority_address()` - returns `slice`, that is authority's address. Authority can revoke SBT.
+
+### Implementation example
+https://github.com/getgems-io/nft-contracts/blob/main/packages/contracts/sources/sbt-item.fc
+
+# Guide
+
+#### Minting
+It can be done using basic NFT collection, SBT should be an item. In mint message additionally uint256 owner's public key should be passed, [after content](https://github.com/getgems-io/nft-contracts/blob/main/packages/contracts/sources/sbt-item.fc#L137). 
+
+#### Changing owner's address
+If you migrated to newer version of wallet and you want to move your SBT to it, you could send transfer to SBT from new wallet with payload:
+```
+pull_ownership#03fdd6c9 query_id:uint64 signature:^(signature:(bits 512)) 
+sbt_nonce:uint64 new_owner:MsgAddress response_destination:MsgAddress 
+custom_payload:(Maybe ^Cell)
+```
+1. To do this, first you need to know current SBT's nonce, you can trigger `get_nonce` method of the SBT contract to get it.
+2. `new_owner` should equals your wallet from which you sends message.
+3. Then you need to sign `sbt_nonce:uint64 new_owner:MsgAddress response_destination:MsgAddress custom_payload:(Maybe ^Cell)` this part of the message and put signature as first reference.
+4. Now you can send this message as internal to SBT and owner's address will be changed to your new wallet's address. 
+
+It is also possible to destroy SBT by setting `new_owner` to null address. After that, owner's address cannot be changed anymore.
+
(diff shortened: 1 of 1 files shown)


text/0083-sbt-standard.md
```diff
@@ -0,0 +1,255 @@
...
+
+SBT implements [NFT standard interface](https://github.com/ton-blockchain/TIPs/issues/62) but `transfer` should always be rejected, `pull_ownership` is used instead.
+
+#### 1. `pull_ownership`
+
+TL-B schema of inbound message:
+```
+pull_ownership#03fdd6c9 query_id:uint64 signature:^(signature:(bits 512)) 
+sbt_nonce:uint64 new_owner:MsgAddress response_destination:MsgAddress 
+custom_payload:(Maybe ^Cell)
```
EmelyanenkoK, line 26: `= InternalMsgBody;`
should be added

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,255 @@
...
+   `excesses#d53276db query_id:uint64 = InternalMsgBody;`
+
+   query_id should be equal with request's `query_id`.
+
+#### 2. `prove_ownership`
+
+TL-B schema of inbound message:
+```
+prove_ownership#38061b82 query_id:uint64 dest:MsgAddress 
+data:^Cell with_content:bool
```
EmelyanenkoK, line 61: `= InternalMsgBody;`
should be added

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,255 @@
...
+`query_id` -  arbitrary request number.
+
+`dest` -  address of the contract to which the ownership of SBT should be proven.
+
+`data` - arbitrary data required by target contract.
+
+`with_content` - if true, SBT's content cell will be included in message to contract.
+
+**Should be rejected if:**
+Sender address is not an owner address.
```
EmelyanenkoK, line 72: Why?

xssnick: Good question! 

Yes, actually I see no reasons to restrict it, we can allow to trigger it to anyone to prove ownership with current owner's address.

xssnick: Updated implementation and description

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,252 @@
...
+
+`owner` - current owner's address.
+
+`data` - data cell passed in `prove_ownership`.
+
+`content` - SBT's content, it is passed if `with_content` was true in `prove_ownership`.
+
+In case when `verify_ownership` was bounced back to SBT, SBT should send message to owner with schema:
+```
+verify_ownership_bounced#81b510c2 query_id:uint64 sbt_id:uint256 owner:MsgAddress
```
EmelyanenkoK, line 91: Does owner need sbt_id, owner, data and content in this case?

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,252 @@
...
+```
+destroy#1f04537a query_id:uint64 = InternalMsgBody;
+```
+`query_id` -  arbitrary request number.
+
+**Should be rejected if:**
+Sender address is not an owner's address.
+
+**Otherwise should do:**
+Set owner's address to null and set public key to 0.
```
EmelyanenkoK, line 107: Do we need response with excess money here back to owner?
Do we want to keep enough balance on SBT to ensure it will not be frozen and destroyed for a long time?

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,252 @@
...
+Sender address is not an authority's address.
+
+**Otherwise should do:**
+Set owner's address to null and set public key to 0.
+
+**GET methods**
+1. `get_public_key()` - returns `int`, that is owner's public key.
+2. `get_nonce()` - returns `int`, which current nonce.
+3. `get_nft_data()` - same as in [NFT standard](https://github.com/ton-blockchain/TIPs/issues/62).
+4. `get_authority_address()` - returns `slice`, that is authority's address. Authority can revoke SBT.
```
EmelyanenkoK, line 127: If authority can not be changed may be it is better to force collection to be authority? Not sure though

xssnick: Then it will require custom collection to work with SBT, with additional revoke method, not sure if it will be good.

text/0083-sbt-standard.md
xssnick: It is like proxy of bounced message, technically i think most of that it is not so useful, but just for similarity with original.

As a use case, sbt_id can be used to confirm that bounce was from real SBT, by calculating address and comparing with sender.

text/0083-sbt-standard.md
xssnick: Good suggestion, I think we can add reserve of 0.05 TON and return excess back, if balance is less than 0.05 throw an error. Do you think it is good approach?

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,252 @@
...
+
+Initially, the design similar to ETH with address-bounded tokens was considered, but because of difference in TON architecture, especially wallet versions, design was reworked. After that, special **killer feature** with ownership proof to contract was added.
+
+- **What is the impact of not doing this?**
+
+Currently, TON have no owner-bounded token standard, so it is a problem to issue tokens that cannot be transferred to 3rd parties. So, if we ignore this or any similar standard that introduces such mechanics, TON could miss some interesting and perspective products.
+
+# Prior art
+
+In ETH - SBT was done as an NFT which could not be transferred between wallets at all, but in TON - architecture is different, and sometimes it is required to update wallet version. This action will also change wallet address but owner will remain the same. Thus, pull ownership method was introduced to change owner's wallet address of SBT.
```
hacker-volodya, line 247: Should include link to the EIP

hacker-volodya:
"Drawbacks" section is missing.

xssnick:
> "Drawbacks" section is missing.

Added

text/0083-sbt-standard.md
xssnick: Done

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,265 @@
...
+sbt_nonce:uint64 new_owner:MsgAddress response_destination:MsgAddress 
+custom_payload:(Maybe ^Cell) = InternalMsgBody;
+```
+`query_id` -  arbitrary request number.
+
+`signature` - signature of the rest part of the message.
+
+`sbt_nonce` - nonce, required for protection of signature replay attacks.
+
+`new_owner` - address of the new owner of the SBT item, should be the same as sender's address.
```
oleganza, line 34: Why don't we use sender's address directly, without duplicating it in the message?

EmelyanenkoK: In your case, imagine situation when we have quite sharded chain and new owner from shard `0b00000` try to `pull_ownership` from sbt deployed in `0b11111`. Message from new owner will migrate to new address some time during which attacker from the same shard may be able to send exactly the same message (signature will be valid since it doesn't include anything specific to new owner) and get sbt. While it is not critical (ownership is still bound to the public key, not the address), it is bad enough to avoid such attacks.

oleganza: Could you clarify what does "address of the new owner [...] should be the same as sender's address" mean? If we need to deal with sharding in some specific way, shouldn't we put this into spec?

oleganza: Ok, i see, this makes signature verification simpler this way. In principle, we should weigh the cost of transmitting extra 32 bytes vs gas spent constructing a cell for signature, but i doubt those extra 32 bytes would make a noticeable difference.

oleganza:
I looked at the spec, I have a few cosmetic suggestions:

### 1. Owner & Wallet

Let's use consistent naming for **owner** and **wallet**. The SBT is designed to be bound to the _owner_, but leaves ability to switch the _wallet_ in case the owner upgrades from v3 to v4 or to v5, while keeping the same private key inside.

Rename pubkey method to clarify the role of the pubkey:

`owner` field
`get_public_key()` -> `get_owner_key()`

Rename "pull ownership" because it sounds like some authority is taking away your ownership. We use it to switch the wallet, but keep the same owner.

`pull_ownership(... new_owner: ...)` -> `switch_wallet(... new_wallet: ...)`

### 2. Nonce vs Seqno

Seems like doing `seqno+=1` like in the wallet code is the simplest replay protection possible, so maybe we should rename `nonce` into `seqno`.

### 3. Unify revoke & destroy?

Maybe the spec could be made simpler if we unify `revoke` and `destroy` as one method `destroy`. If we want to tag things in the explorer or API, we could simply check if the msg sender is the SBT's wallet or the authority and mark the transaction accordingly.

oleganza:
From the UX perspective the SBT always belongs to the wallet, so `switch_wallet` method (currently `pull_ownership`) is going to be used within some migration flow, not as a general purpose "transfer" UI. This also means that the wallet is not concerned with transferring the privkey of SBT to any other location and securing it along the way.

xssnick:
> I looked at the spec, I have a few cosmetic suggestions:
> 
> ### 1. Owner & Wallet
> Let's use consistent naming for **owner** and **wallet**. The SBT is designed to be bound to the _owner_, but leaves ability to switch the _wallet_ in case the owner upgrades from v3 to v4 or to v5, while keeping the same private key inside.
> 
> Rename pubkey method to clarify the role of the pubkey:
> 
> `owner` field `get_public_key()` -> `get_owner_key()`
> 
> Rename "pull ownership" because it sounds like some authority is taking away your ownership. We use it to switch the wallet, but keep the same owner.
> 
> `pull_ownership(... new_owner: ...)` -> `switch_wallet(... new_wallet: ...)`
> 
> ### 2. Nonce vs Seqno
> Seems like doing `seqno+=1` like in the wallet code is the simplest replay protection possible, so maybe we should rename `nonce` into `seqno`.
> 
> ### 3. Unify revoke & destroy?
> Maybe the spec could be made simpler if we unify `revoke` and `destroy` as one method `destroy`. If we want to tag things in the explorer or API, we could simply check if the msg sender is the SBT's wallet or the authority and mark the transaction accordingly.

1. Yea, agreed with switch_wallet, it looks better, will update. But not sure about changing get_public_key to get_owner_key, because currently interface is same with wallet’s method, it allows to universally get the owner’s key of both entities.
2. Nonce was used here because with seqno we have a low chance of replay attack, if 2 sbts of the same owner of same collection are connected to one key. If seqno is same, attacker can repeat the request on second sbt and change wallet address to same as on first one. It is not so critical, since it is very rare case, but nonce resolves it and looks better here.
3. I think revoke and destroy are different processes and in some cases they can contain additional logic, so imo it is better to keep them as separate operations.

oleganza:
Another note: conceptually, we need to put as little logic in the contract as possible to enforce relationship between the designated parties. In our case the parties are: "the owner" (identified by a fixed pubkey) and "the authority" (maybe rename this to "issuer"?). 

Do we really need a concept of a "wallet" here at all? If we don't then:

1. We do not need to have `switch_wallet` logic at all. When user migrates the wallet from v5 to v7 nothing has to happen to SBP. We can move that signature verification into the `prove_ownership` method instead.
2. Wallets need to have a special support for SBTs anyway (to prove ownership and display them distinctly from NFTs because SBTs can't be transferred, to support token-specific features), so they might as well pick them up by the pubkey instead of the NFT ownership API.

NFTItem interface is cool because you control that thing with an arbitrary address (which could be plain wallet, multisig, etc.), but SBT is not really controlled by the wallet, but explicitly controlled by a fixed pubkey.

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,265 @@
...
+   `excesses#d53276db query_id:uint64 = InternalMsgBody;`
+
+   query_id should be equal with request's `query_id`.
+
+#### 2. `prove_ownership`
+
+TL-B schema of inbound message:
+```
+prove_ownership#04ded148 query_id:uint64 dest:MsgAddress 
+forward_payload:^Cell with_content:Bool = InternalMsgBody;
```
oleganza, line 61:
```suggestion
data:^Cell with_content:Bool = InternalMsgBody;
```

?

text/0083-sbt-standard.md
```diff
@@ -0,0 +1,265 @@
...
+
+# Specification
+
+SBT implements [NFT standard interface](https://github.com/ton-blockchain/TIPs/issues/62) but `transfer` should always be rejected, `pull_ownership` is used instead.
+
+#### 1. `pull_ownership`
+
+TL-B schema of inbound message:
+```
+pull_ownership#08496845 query_id:uint64 signature:^(bits 512)
```
oleganza, line 24: why `query_id` is before the signature? Shouldn't it be covered by the signature and it'd be easier to verify it if sig comes first?

xssnick:
After additional discussions standard was split to this one and NFT Ownership, and simplified.

Implementation and description was changed.

oleganza:
I am pretty happy with the concept of making SBT a ≈subset of NFTs. The definition "SBT = NFT where only the minter/issuer/authority can `transfer`" seems like the minimal possible one. From the perspective of the ecosystem that's enough. 

But for the security of the issuer they need to verify that the recipient's wallet is a regular wallet v3/v4 so they don't accidentally issue SBT to a transferrable "holding contract" that would effectively "wrap" SBT into a transferrable NFT and hamper the bonding property.

Naltox (approved):

EmelyanenkoK:
LGTM

dvlkv (approved):
LGTM.

tvorogme:
Great work, LGTM

TrueCarry (approved):

hacker-volodya:
This TEP enters its Final Comment Period. It ends in 10 calendar days, 11.10.2022.

hacker-volodya:
The Final Comment Period is complete with a disposition to **merge**. Thanks to the authors of TEP, @xssnick, @Naltox, @EmelyanenkoK, @oleganza, and to all who took part in the discussion. This pull request will be merged now.

MrG9090 (approved):

PR #86, opened 2022-09-03
TEP: SFT Standard

ivklim-ton-play:
Semi-fungible tokens(SFT) standard is needed in the case when you need to issue a large number of identical NFTs, but the serial number is not important. For example, a membership card to access the site or the same items for games (100 red t-shirts, 100 yellow shorts, etc.).

SFT represents ownership of a copy of a unique asset in a collection.


EmelyanenkoK:
If I understand it correctly, basically we have Collection which mints Jetton-Minters which mints Jettons.
The difference from standard Collection,Minters and Jettons is in how we retrieve Metadata: for usual jettons it is stored in minter, while here it is calculated for minters the same way it is calculated for nft-items.
To notify the difference it proposes to rename get_methods:
- `get_sft_wallet_data` instead of `get_walled_data` for jetton-wallets
- `get_sft_data` instead of `get_jetton_data` for jetton-minters
- `get_sft_collection_data`,`get_sft_minter_content` and `get_sft_minter_address_by_index` instead of `get_collection_data`,`get_nft_content` and `get_nft_address_by_index`

@mr-tron @ex3ndr Share your thoughts on that please

ex3ndr:
Adding new asset types is unnecessarily burden on everyone in the ecosystem. 

What's this token solves that couldn't be solved with just jettons?

ivklim-ton-play:
> If I understand it correctly, basically we have Collection which mints Jetton-Minters which mints Jettons. The difference from standard Collection,Minters and Jettons is in how we retrieve Metadata: for usual jettons it is stored in minter, while here it is calculated for minters the same way it is calculated for nft-items. To notify the difference it proposes to rename get_methods:
> 
> * `get_sft_wallet_data` instead of `get_walled_data` for jetton-wallets
> * `get_sft_data` instead of `get_jetton_data` for jetton-minters
> * `get_sft_collection_data`,`get_sft_minter_content` and `get_sft_minter_address_by_index` instead of `get_collection_data`,`get_nft_content` and `get_nft_address_by_index`
> 
> @mr-tron @ex3ndr Share your thoughts on that please

Thank you very much for the analysis of the proposed standard. I think you are describing the proposed differences from Jettons and NFT as close and concise as possible.
I tried to keep exactly the idea of ​​​​NFT and Jettons, so that new developers would more easily understand the SFT standard, since it is their combination (which was described below in one sentence https://github.com/ton-blockchain/token-contract ). Most likely, it will be easier for current services to add this standard to themselves, since they have already worked with NFTs and Jettons.

ivklim-ton-play:
> Adding new asset types is unnecessarily burden on everyone in the ecosystem.
> 
> What's this token solves that couldn't be solved with just jettons?

Thank you very much for your question. I think Jettons is a great representation of tokens.

But the advantage over just a bunch of individual tokens is in authentication. A user/service can only verify the root contract and immediately start trusting all child contracts (even those that have not yet appeared). In the system, there are just adjacent tokens - each token must be separately entered into whitelists conditionally (and when new tokens appear in the future, too).

Any third-party service looking at the contract must understand how to correctly display the content. For example, it is important to display items in the game exactly as NFTs, and not coins. For users, confusion is introduced (in the wallet) when 100 T-shirts are displayed as Jettons (but traded in the NFT marketplace) and Jettons of the game itself are nearby (which are further traded somewhere on the DEX).

ivklim-ton-play:
I think 3 main standards for EVM are recognized in the world for a reason. They are used by a lot of people.
 
ERC-20 (in Ton it is Jettons) it is well suited for trading on the stock exchange. In this format, you can represent a currency or securities in the form of tokens of some service.
ERC-721 (in Ton it is NFT) some unique entities are well represented by this format. From life it is similar to paintings, but now you can imagine anything unique.
But ERC-1155 is very lacking in Ton. They are not traded on the exchanges as a currency, but they are not absolutely unique. It's more like branded clothes that were made at the factory. This standard is great for games and services that require the same cards without a unique design. So what does SFT look like.

There is another story when teams want to create cross-chain projects. Most of the popular blockchains already support the ERC-1155 analogue. When game developers come to Ton, their first question is "Is there an analogue of ERC-1155?". Even if the team switches to only Jettons, they will face the problem that all of them (ERC-1155) will be displayed as coins + they are not related to each other. In order to somehow display them correctly and combine them into something related, you will have to contact third-party services, which may not have such functionality at all. This can cause the blockchain to lose potentially good projects and teams. It seems to me that it is worth combining good practices from other projects and making them even better, or at least not ignoring them.

ex3ndr:
Sorry, but all this comments are NOT the answers to a simple question - why it requires breaking(!) changes?

The main selling point for ERC-1155 is NOT the need to deploy numerous contracts for each type of asset, but this proposal does nothing to prevent this. You must deploy a new contract each time you introduce a new "collection".

The appropriate implementation for TON would build a new "jetton/nft"-like a contract that could hold multiple different assets with different IDs and therefore not require a nearly infinite number of contracts (for worst case).

Correct me if I am wrong, but I don't see how this PR solves the most critical issue that ERC-1155 solves.

ivklim-ton-play:
I sincerely try to contribute something useful to Ton and am ready to answer questions. Also open to discussion.

> "why does it require breaking(!) changes?"

Unfortunately, the question is not clear without context. The proposed standard only combines the best practices of NFTs and Jettons. I'm definitely not suggesting that we remove the NFT and Jettons standards.

> "The main selling point for ERC-1155 is NOT the need to deploy numerous contracts for each type of asset, but this proposal does nothing to prevent this. You must deploy a new contract each time you introduce a new "collection.""

This comment is more about the Ton feature. In the EVM, the allocated memory in the Store is paid once when you occupy it (and you are also reimbursed for some of the ether when you free it). In Ton, the allocated memory in the contract is paid for every time someone uses the contract (i.e. each subsequent user pays for the memory that the contract occupies). This imposes a feature on the contract when dealing with memory. As soon as we want to enter some HashMap(Dictionary), then the contract memory will be dynamic and it is difficult to predict how much gas is needed (since Ton is also asynchronous). In this case, in place of HashMap, we deploy data, or rather, even a whole contract with functions separately.
It works in exactly the same way with jetton-wallet. Every time we deploy a new wallet for each user. We cannot afford to keep a table of users, as it is in ERC-20.
Indeed, for each group, I propose to deploy a new sft-minter contract, which will refer to one collection. Unfortunately, due to the peculiarities of Ton, we cannot implement the absolute equivalent of ERC-1155, but with this approach, we need to abandon Jettons as well, which I think will only move us back from improving the network.

> The appropriate implementation for TON would build a new "jetton/nft"-like a contract that could hold multiple different assets with different IDs and therefore not require a nearly infinite number of contracts (for worst case).

If I understand correctly, then the question is:

Do we really need an infinite number of sft-minter.

This is a question that can really be discussed. Since most often there is no collection of 10000000 id.
One could imagine that we still do a hashmap inside the contract, but then each such SFT will cost a different gas, in order to interact with it, each service needs to ask each time if the size has changed and how much gas is now. Prescribing immutability in the standard is too strict a requirement. As a result, the problem of gas unpredictability arises again.

> Correct me if I am wrong, but I don't see how this PR solves the most critical issue that ERC-1155 solves.

The main thing to remember is that we do not solve the problems of ERC-1155 and do not try to duplicate it. ERC-1155 is a good example of how several unrelated tokens are combined into one collection under different ids, and tokens under the same id can be transferred at once in one transaction in several pieces.
These are the important things we need right now. It is important that third-party services for get methods can correctly display all the necessary data.
At least 3 important things
- transfer of several tokens in one operation and the same cost
- combining several types of tokens (and their data) under one collection and checking belonging to this collection we can do
- services will have a single interface for reading and displaying user tokens 

Thank you very much for your questions and interest

> Sorry, but all this comments are NOT the answers to a simple question - why it requires breaking(!) changes?
> 
> The main selling point for ERC-1155 is NOT the need to deploy numerous contracts for each type of asset, but this proposal does nothing to prevent this. You must deploy a new contract each time you introduce a new "collection".
> 
> The appropriate implementation for TON would build a new "jetton/nft"-like a contract that could hold multiple different assets with different IDs and therefore not require a nearly infinite number of contracts (for worst case).
> 
> Correct me if I am wrong, but I don't see how this PR solves the most critical issue that ERC-1155 solves.

ex3ndr:
> Unfortunately, the question is not clear without context. The proposed standard only combines the best practices of NFTs and Jettons. I'm definitely not suggesting that we remove the NFT and Jettons standards.

You are proposing a new token that is just a combination of NFT and Jettons, but adds few methods making them incompatible with jettons and NFTs. This differences are non-significant to support completely different token from apps point of view. You can just add few fields to metadata to NFT and Jettons and we will alter behaviour of apps for such jettons/nft instead of implementing them from scratch (since they are not compatible).

> his comment is more about the Ton feature. In the EVM, the allocated memory in the Store is paid once when you occupy it (and you are also reimbursed for some of the ether when you free it). In Ton, the allocated memory in the contract is paid for every time someone uses the contract (i.e. each subsequent user pays for the memory that the contract occupies). This imposes a feature on the contract when dealing with memory. As soon as we want to enter some HashMap(Dictionary), then the contract memory will be dynamic and it is difficult to predict how much gas is needed (since Ton is also asynchronous). In this case, in place of HashMap, we deploy data, or rather, even a whole contract with functions separately.
It works in exactly the same way with jetton-wallet. Every time we deploy a new wallet for each user. We cannot afford to keep a table of users, as it is in ERC-20.

This is completely NOT what I said, I haven't proposed to have a hashmap of users, but a hashmap of assets in each token-wallet. This is exactly how it works in ETH. You can generate IDs of assets dynamically and resolve them completely offchain. 

For example i have an asset with very high variability in its features and parameters. For example, some hats. The hat could have multiple forms multiplied by multiple colors multiplied by the size of a hat. Having five different forms, ten colors, and four sizes, we are looking at 5 * 10 * 4 = 200 different assets.

ERC-1155 solves this, but your proposal forces the developer to create 200 minters, and each user would have 200 associated asset addresses.  Adding new color? This would make it 20+ more collections and addresses. Now imagine how much work it takes for apps, indexers, and stuff to track all of them. Virtually for each asset, we will need to track its balances on apps this means much more requests everywhere. This is just pure insanity.

ERC-1155 alternative in TON would be a single wallet for each user that holds all related assets and is very easy to manage and work.

> Prescribing immutability in the standard is too strict a requirement. As a result, the problem of gas unpredictability arises again.

All gas operations could be predictable and always under some reasonable threshold.

ivklim-ton-play:
> > Unfortunately, the question is not clear without context. The proposed standard only combines the best practices of NFTs and Jettons. I'm definitely not suggesting that we remove the NFT and Jettons standards.
> 
> You are proposing a new token that is just a combination of NFT and Jettons, but adds few methods making them incompatible with jettons and NFTs. This differences are non-significant to support completely different token from apps point of view. You can just add few fields to metadata to NFT and Jettons and we will alter behaviour of apps for such jettons/nft instead of implementing them from scratch (since they are not compatible).
> 
> > his comment is more about the Ton feature. In the EVM, the allocated memory in the Store is paid once when you occupy it (and you are also reimbursed for some of the ether when you free it). In Ton, the allocated memory in the contract is paid for every time someone uses the contract (i.e. each subsequent user pays for the memory that the contract occupies). This imposes a feature on the contract when dealing with memory. As soon as we want to enter some HashMap(Dictionary), then the contract memory will be dynamic and it is difficult to predict how much gas is needed (since Ton is also asynchronous). In this case, in place of HashMap, we deploy data, or rather, even a whole contract with functions separately.
> > It works in exactly the same way with jetton-wallet. Every time we deploy a new wallet for each user. We cannot afford to keep a table of users, as it is in ERC-20.
> 
> This is completely NOT what I said, I haven't proposed to have a hashmap of users, but a hashmap of assets in each token-wallet. This is exactly how it works in ETH. You can generate IDs of assets dynamically and resolve them completely offchain.
> 
> For example i have an asset with very high variability in its features and parameters. For example, some hats. The hat could have multiple forms multiplied by multiple colors multiplied by the size of a hat. Having five different forms, ten colors, and four sizes, we are looking at 5 * 10 * 4 = 200 different assets.
> 
> ERC-1155 solves this, but your proposal forces the developer to create 200 minters, and each user would have 200 associated asset addresses. Adding new color? This would make it 20+ more collections and addresses. Now imagine how much work it takes for apps, indexers, and stuff to track all of them. Virtually for each asset, we will need to track its balances on apps this means much more requests everywhere. This is just pure insanity.
> 
> ERC-1155 alternative in TON would be a single wallet for each user that holds all related assets and is very easy to manage and work.
> 
> > Prescribing immutability in the standard is too strict a requirement. As a result, the problem of gas unpredictability arises again.
> 
> All gas operations could be predictable and always under some reasonable threshold.

ivklim-ton-play:
> You are proposing a new token that is just a combination of NFT and Jettons, but adds few methods making them incompatible with jettons and NFTs. This differences are non-significant to support completely different token from apps point of view. You can just add few fields to metadata to NFT and Jettons and we will alter behaviour of apps for such jettons/nft instead of implementing them from scratch (since they are not compatible).

Today I prepared [sample code](https://github.com/ivklim-ton-play/ton-SFT) and a [description of the new version standard.](https://github.com/ivklim-ton-play/TEPs/blob/master/text/0084-sft-standart.md) 
Thank you very much for your comment.

I received feedback from a few more people and updated the version of the standard today.

The new version of the proposal to go from the Jettons. Expand the tokens and unite it all under the collection from NFT.

**sft-wallet** is the same **jetton-wallet**

**sft-minter** is that **jetton-master**, but it has transfer and get_static_data methods from NFT Iten, but get methods only from jettons.
Additionally, it has a separate method to get information about the get_sft_collection_data collection.

**sft-collection** is the same **nft-collection**

it is proposed to add the sft minter object to the sft-collection and sft-minter metadata. Object because it will be easier to expand it in the future.

If there are any additional technical features that are worth learning, I will be glad to hear.

I'm still thinking whether it's worth adding some method to the SFT Collection in order to somehow distinguish it from the NFT Collection. Need advice

> This is completely NOT what I said, I haven't proposed to have a hashmap of users, but a hashmap of assets in each token-wallet. This is exactly how it works in ETH. You can generate IDs of assets dynamically and resolve them completely offchain.

> For example i have an asset with very high variability in its features and parameters. For example, some hats. The hat could have multiple forms multiplied by multiple colors multiplied by the size of a hat. Having five different forms, ten colors, and four sizes, we are looking at 5 * 10 * 4 = 200 different assets.

> ERC-1155 solves this, but your proposal forces the developer to create 200 minters, and each user would have 200 associated asset addresses. Adding new color? This would make it 20+ more collections and addresses. Now imagine how much work it takes for apps, indexers, and stuff to track all of them. Virtually for each asset, we will need to track its balances on apps this means much more requests everywhere. This is just pure insanity.

> ERC-1155 alternative in TON would be a single wallet for each user that holds all related assets and is very easy to manage and work.

Unfortunately we will have to deploy 200 different Jettons anyway for game for example. But I would like to combine them into a collection.

Yes, it would be great to store in one wallet. But unfortunately, I don't see such an opportunity for Ton yet. I can hardly imagine that each service will store some kind of merkle tree off-chain to make confirmation.

ivklim-ton-play:
Based on community feedback, since an empty object can mean false, it is suggested to use the text "sft": "true" in the metadata instead of "sft": {}.

mr-tron:
LGTM
@EmelyanenkoK @ex3ndr

ivklim-ton-play:
Based on community feedback, I propose to remove the mandatory requirement transfer for  SFT minter.

oleganza:
Here's my personal summary based on coversations with the authors of this proposal:

First, let's not use "SFT" as a term. We already have low-level building blocks called "FT" and "NFT" (for fungible tokens aka "jettons"; and atomic non-fungible items aka "nfts"). Whenever you need something fungible, you use jetton.

Then, the actual requirements are two-fold and orthogonal to each other:

1. We want some tokens to be organized in collections, like NFT items are organized in collections. E.g. in games you may have resources emitted by the same game engine. Or in banking you may have multiple currencies issued by different banks.

2. We want some jettons to be rendered in a more fancy manner than a plain list (irrespective of whether they are part of a collection or not), but more like NFT cards. E.g. game resources have bigger, juicier images.

3. The rendering in (2) may have more specifics: e.g. you may also show a portion of total limited supply in multiple styles. E.g. "12 of 100 gems" or "25% of wood".

There was also another requirement about grouping items to permit atomic and cheaper transfers of multiple items. This is no longer a requirement, however we generated a few ideas on how to make universal jetton aggregator or one-time bundles for use in, say, liquidity pools.

N3koSempai:
I support the use of a new contract.
But why make a contract more complex for two different use cases and increase the difficulties and possibilities of error. Also adding complexity to anyone who wants to launch this type of token as it does not follow the standards currently used in other blockchains. Instead of using a specific contract for a use case you have to modify a contract making it more ambiguous and adding points of failure. 
When you can simply separate it into different contracts which is the solution currently considered state of the art in all other blockchains.

example case of use :
https://www.coindesk.com/tech/2021/08/17/what-is-a-semi-fungible-crypto-token/

ivklim-ton-play:
I'm closing this thread so no one gets confused.

Here is a link to a pull request that is specifically about rendering
https://github.com/ton-blockchain/TEPs/pull/98

There will be a separate pull request for Jetton Collections. The term SFT will not be used there, so as not to introduce new terms. The standard will propose to combine Jettons into collections or even introduce the general concept of a collection standart on the blockchain.

brainpicture:
Enhancement listed above solved problem of informing the wallets that specific jettons could be shown in the wallets differently, (more like NFT) without limiting the compatibility with default Jettons spec and without limiting the extensibility.
Therefor I vote to accept this draft!

gagapuh:
So, friends, can I use SFT in the TON blockchain for multiple instances, analogous to ERC-1155? As far as I understand, there is no native support for this in TON, right?

I need to issue SFT, meaning unique items, but in quantities greater than one. In other words, multiple instances.

PR #88, opened 2022-09-07
TEP: TVM update

EmelyanenkoK:
This proposal is still work in progress and not scheduled even for testing in public testnet: we are figuring out what else updates in TVM can be needed in the future.

### DIFF
--- a/text/0088-tvm-update.md
+++ b/text/0088-tvm-update.md
@@ -1,4 +1,4 @@
-- **TEP**: [87](https://github.com/ton-blockchain/TEPs/pull/87)
+- **TEP**: [88](https://github.com/ton-blockchain/TEPs/pull/87)
 - **title**: TVM update
 - **status**: Draft
 - **type**: Core
--- a/text/0087-tvm-update.md
+++ b/text/0087-tvm-update.md
@@ -1,0 +1,142 @@
+- **TEP**: [87](https://github.com/ton-blockchain/TEPs/pull/87)
+- **title**: TVM update
+- **status**: Draft
+- **type**: Core
+- **authors**: [@SpyCheese](https://github.com/SpyCheese) [@EmelyanenkoK](https://github.com/EmelyanenkoK) 
+- **created**: 07.07.2022
+- **replaces**: -
+- **replaced by**: -
+
+# Summary
+
+This proposal suggests to
+- add new op-codes to TVM related to block data retrieving, hashing, cryptographic signing and message sending
+- include into `c7` register of TVM smart-contract' own code, additional data on transaction and information about masterchain blocks
+ 
+# Motivation
+
+Current TVM implementation makes it hard to implement some common thing. Proposal aims to make it easier and cheaper to:
+- hash big chanks of data (including non-sha256 hashes)
+- work with Bitcoin/Ethereum compatible signtatures
+- prove TON onchain data
+- calculate fee for sending messages
+
+# Guide
+
+Draft implementation is available [here](https://github.com/SpyCheese/ton/commits/new-opcodes).
+
+This proposal suggests to 
+1. Extend **c7** tuple from 10 to 14 elements:
+   - **10**: code of the smart contract.
+   - **11**: value of the incoming message.
+   - **12**: fees collected in the storage phase.
+   - **13**: information about previous blocks.
+2. Add the following TVM op-codes:
+   - **c7** primitives: `MYCODE`, `INCOMINGVALUE`, `STORAGEFEES`, `PREVBLOCKSINFOTUPLE`
+   - Block primitives: `PREVMCBLOCKS`, `PREVKEYBLOCK`
+   - Hash primitives: `HASHSTART`, `HASHEND`, `HASHENDST`, `HASHINFO`, `HASHAPPU`, `HASHAPPI`, `HASHAPPS`, `HASHAPPB`
+   - Cryptography primitives: `ECRECOVER`
+   - Message primitives: `SENDMSG`
+
+3. Change sending fee calculation: deduct fee related to message estimation even in the case of failed sending.
+
+# Specification
+
+## c7 update
+**10th** element of **c7** tuple contains cell with executed smart-contract code (from init_state of incoming message if applicable)
+
+**11th** element of **c7** tuple contains value of incoming message. It is a tuple with two elements: incoming value (same value which is put on stack as 4th element) and a dictionary of extra currencies. If not applicable, the value is `[0, null]`.
+
+**12th** element of **c7** tuple contains fee debited in storage phase.
+
+**13th** elements of **c7** contains of tuple with two elements: `last_mc_blocks` and `prev_key_block`.
+- `last_mc_blocks` is a tuple of up to 16 elements which contains previous most recent masterchain blocks info (ordered by seqno). Each block is represented as tuple of 5 integers: `(workchain, shard, seqno, root_hash, file_hash)`
+- `prev_key_block` is a tuple of 5 integers `(workchain, shard, seqno, root_hash, file_hash)` corresponding to most recent key block.
+
+## TVM update
+### New stack type
+This proposal suggests to introduce new type in TVM `Hasher`: it is opaque type used to consequently hash data chunks. This type behave the same way as others and can be considered immutable: `HASHAPP*` ops (described below) "consume" old hasher and return updated one (that way if old hasher was copied, it's copies are completely independent). This object is not serializable and thus can not be returned from TVM (for instance as result of get-method).
+### New opcodes
+#### c7 and block ops
+`MYCODE` - `0xF82A` - ` - c` - return cell with currently executed code. Equivalent to `10 GETPARAM`.
+
+`INCOMINGVALUE` - `0xF82B` - ` - i` - return integer with TON amount in incoming message (or zero if not applicable). Equivalent to `11 GETPARAM`.
+
+`STORAGEFEES` - `0xF82C` - ` - i` - return integer value of fee in nanoTONs debited in storage phase. Equivalent to `12 GETPARAM`.
+
+`PREVBLOCKSINFOTUPLE` - `0xF82D` - ` - t` - return tuple with previous blocks info. Equivalent to `13 GETPARAM`.
+
+`PREVMCBLOCKS`- `0xF83400` - ` - t` - return tuple with previous masterchain blocks info. Equivalent to `13 GETPARAM; FIRST`
+
+`PREVKEYBLOCK`- `0xF83401` - ` - t` - return tuple with previous keyblock id. Equivalent to `13 GETPARAM; SECOND`
+
+#### Hash primitives
+`HASHSTART_SHA256` - `0xF90300` - ` - h` - create `sha256` hasher object.
+
+`HASHSTART_SHA512` - `0xF90301` - ` - h` - create `sha512` hasher object.
+
+`HASHSTART_BLAKE2B` - `0xF90302` - ` - h` - create `blake2b` hasher object.
+
+`HASHSTART_KECCAK256` - `0xF90303` - ` - h` - create `keccak256` hasher object.
+
+`HASHSTART_KECCAK512` - `0xF90304` - ` - h` - create `keccak512` hasher object.
+
+
+`HASHEND` - `0xF904` - `h - ... ` - calculate hash from hasher and put in on the stack. If size of the hash does not exceed 256 bits, it is returned as a 256-bit unsigned integer (e.g. sha256). Otherwise it is returned as a tuple of 256-bit integers (e.g. sha512 - tuple of two integers). If the bit length of data in hasher is not divisible by eight, throws a cell underflow exception.
+
+`HASHENDST` - `0xF905` - `b h - b' ` - calculate hash from hasher and store it into the slice.
+
+
+`HASHINFO` - `0xF906` - `h - i` - return hash type of hasher (`sha256` - `0`, `sha512` - `1`, `blake2b` - `2`, `keccak256` - `3`, `keccak512` - `4`).
+
+
+`HASHAPPU` - `0xF907xx` - `h i - h'` - serialize unsigned integer in `xx+1` number of bits and put it into hasher.
+
+`HASHAPPI` - `0xF907xx` - `h i - h'` - serialize signed integer in `xx+1` number of bits and put it into hasher.
+
+`HASHAPPS` - `0xF909` - `h s - h'` - put data bits from slice to hasher.
+
+`HASHAPPB` - `0xF909` - `h b - h'` - put data bits from builder to hasher.
+
+#### Cryptography primitives
+
+`ECRECOVER` - `0xF912` - `m v r s - i i i -1 or 0` - get secp256k1 signature as three integers `v`(8 bit), `r`(256 bit) and `s`(256 bit) and signed message hash `m` as 256 bit integer, decrypts the public key using elliptic curve DSA recovery mechanism and returns pubkey curve point (header-byte integer and two 256-bit integers for coordinates) and `-1` if signature is valid and `0` if not.
+
+#### Message primitives
+`SENDMSG` - `0xFB08` - `c m - i` - Sends a raw message contained in Cell `c` with mode `m` (the same way as `SENDRAWMSG`) and returns approximate action fee. The difference in mode interpreation with `SENDRAWMSG` is as following: `+1024` is used to calculate fee only (doesn't create output action), `+64` uses whole value of incoming value (since exact value can not be determined until the end of Computation phase), `+128` uses whole balance of account (since exact value can not be determined until the end of Computation phase).
+
+### Change in message fee
+It is suggested to deduct fee related to calculation of `fwd_fee` (cell graph traversal) in the case unsuccessfull sending. The fine of visiting cell in message that is not sent is proposed to be 1/4 of msg\_cell\_price.
+
+## Update activation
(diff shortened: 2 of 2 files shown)


text/0088-tvm-update.md
```diff
@@ -0,0 +1,142 @@
...
+
+**12th** element of **c7** tuple contains fee debited in storage phase.
+
+**13th** elements of **c7** contains of tuple with two elements: `last_mc_blocks` and `prev_key_block`.
+- `last_mc_blocks` is a tuple of up to 16 elements which contains previous most recent masterchain blocks info (ordered by seqno). Each block is represented as tuple of 5 integers: `(workchain, shard, seqno, root_hash, file_hash)`
+- `prev_key_block` is a tuple of 5 integers `(workchain, shard, seqno, root_hash, file_hash)` corresponding to most recent key block.
+
+## TVM update
+### New stack type
+This proposal suggests to introduce new type in TVM `Hasher`: it is opaque type used to consequently hash data chunks. This type behave the same way as others and can be considered immutable: `HASHAPP*` ops (described below) "consume" old hasher and return updated one (that way if old hasher was copied, it's copies are completely independent). This object is not serializable and thus can not be returned from TVM (for instance as result of get-method).
```
andreypfau, line 58: I think it's better to use the name "Digest" as in popular languages like Java or GoLang, because this is a more abstract name for hashing/checksum calculation algorithms.

xssnick:
I like most of the new features, but not sure about those:

#### Hasher
I see new `Hasher` object as a not a so good idea. IMO ideology of TON is to keep everything as much universal as
possible, e.g everything is a cell, wallets = smart contracts, etc. Maybe we can implement other hashes in the same way
as 'HASHCU 'HASHSU works? 

I don't actually think that it will be less efficient, I think in the most of use cases we will already have data cell/slice to hash as input.

#### INCOMINGVALUE
Currently we have it on stack when processing internal message, and it is ok, i don't think we need to move it to `c7` at least because we have it only for internal messages, stack looks pretty fine. Less universal and not so useful imo.

text/0088-tvm-update.md
xssnick, line 58: I see new Hasher object as a not a so good idea. 

IMO ideology of TON is to keep everything as much universal as
possible, e.g everything is a cell, wallets = smart contracts, etc. 

Maybe we can implement other hashes in the same way
as 'HASHCU 'HASHSU works?

I don't actually think that it will be less efficient, I think in the most of use cases we will already have data cell/slice to hash as input.

text/0088-tvm-update.md
```diff
@@ -0,0 +1,142 @@
...
+
+`HASHSTART_KECCAK512` - `0xF90304` - ` - h` - create `keccak512` hasher object.
+
+
+`HASHEND` - `0xF904` - `h - ... ` - calculate hash from hasher and put in on the stack. If size of the hash does not exceed 256 bits, it is returned as a 256-bit unsigned integer (e.g. sha256). Otherwise it is returned as a tuple of 256-bit integers (e.g. sha512 - tuple of two integers). If the bit length of data in hasher is not divisible by eight, throws a cell underflow exception.
+
+`HASHENDST` - `0xF905` - `b h - b' ` - calculate hash from hasher and store it into the slice.
+
+
+`HASHINFO` - `0xF906` - `h - i` - return hash type of hasher (`sha256` - `0`, `sha512` - `1`, `blake2b` - `2`, `keccak256` - `3`, `keccak512` - `4`).
```
andreypfau, line 90: I think that hardcoding magic numbers is always a bad practice. There are many algorithms with different names and for sure in the future the list of them will be replenished. If at this stage it is assumed that the ID for the hasher will be chosen purely for "historical reasons", then for many it will cause misunderstanding. I suggest using the same function instead of ordinal numbers as for determining the get-methods ID in smart contracts

text/0088-tvm-update.md
```diff
@@ -0,0 +1,142 @@
...
+- calculate fee for sending messages
+
+# Guide
+
+Draft implementation is available [here](https://github.com/SpyCheese/ton/commits/new-opcodes).
+
+This proposal suggests to 
+1. Extend **c7** tuple from 10 to 14 elements:
+   - **10**: code of the smart contract.
+   - **11**: value of the incoming message.
```
xssnick, line 31: Currently we have it on stack when processing internal message, and it is ok, i don't think we need to move it to `c7` at least because we have it only for internal messages, stack looks pretty fine. Less universal and not so useful imo.

text/0088-tvm-update.md
xssnick: Also, I have an alternative approach which will be same efficent as Hasher, but without introducing new type.

Lets imagine we have opcode `HASH_SHA512` which will accept number of stack elements to hash as first stack element (`s0`).

So if we want to hash 1 uint256 and 1 slice, currently we will do something like:
```
HASHSTART_SHA512
PUSHINT 7777777777
HASHAPPU
PUSHSLICE {001010010101}
HASHAPPS
HASHEND
```
(stack ops are omitted)

But we can do it without hasher type, like:
```
PUSHSLICE {001010010101}
PUSHINT 7777777777
PUSH 2
HASH_SHA512
```
So we read first stack value, it is 2, it means that we need to read 2 more values and hash them, in reverse order. 

This way i think it will be much more clear, and keep architecture cleaner.

EmelyanenkoK: Not sure which "universality" is broken here.
We are looking for the methods which can be used to hash large chunks of data, like 100kB. Approach with `hasher` objects allows to copy partially hashed data and cheaper get digest of data with the same prefix (not sure that we practicaly need it though).

xssnick: By universality I mean that we introducing new type which cannot be returned and serialized to cell. 

Not sure too, where we can use the feature with copying hasher during single contract call, since it cannot be saved and reused later. 

I could propose 2 options here:
1. Do it without hasher object and without copy feature, HASH_SHA512 approach proposed 1 message before, looks  more efficent and smaller (less stack ops needed).
2. Add cell [de]serialization to hasher, then in could be saved, returned and accepted as input, this way prefix feature has more potential I think. And it is actually not a problem to serialize mid state of hasher and unpack later.

EmelyanenkoK: It looks like "digest" refers to result of hashing, here we use `Hasher` for intermediate object with some internal state.

text/0088-tvm-update.md
```diff
@@ -0,0 +1,142 @@
...
+   - **12**: fees collected in the storage phase.
```
ex3ndr, line 32: Can we just adjust the balance by this fee?

text/0088-tvm-update.md
ex3ndr: Isn't this state could be implemented via a generic slice?

ex3ndr: > Not sure which "universality" is broken here.
> We are looking for the methods which can be used to hash large chunks of data, like 100kB. Approach with `hasher` objects allows to copy partially hashed data and cheaper get digest of data with the same prefix (not sure that we practically need it though).

how you can hash 100kb data on a blockchain? It requires much more gas that possible.

text/0088-tvm-update.md
EmelyanenkoK: Not sure what you mean here. Contract balance which can be retrieved by `get_balance()` (and stored in 7th element of c7) is already balance after storage fee deducted. Here we want to give contract an option to account storage fee if necessary.

text/0088-tvm-update.md
EmelyanenkoK: Well, any state can be can be represented as some state. The question do we really need it?

EmelyanenkoK: >how you can hash 100kb data on a blockchain? It requires much more gas that possible
Currently yes, but it doesn't look like intended behavior: things which are cheap to nodes (and hashing is cheap) should not be prohibitively expensive for contracts.

EmelyanenkoK:
Probably it is a good idea to add additional opcodes:
- GLOBALID (can be added to c7), this will allow to implement cross-chain replay protection
- SETBOUNCEONACTIONPHASEFAIL - set/unset transaction executor mode that will bounce on ActionPhase fail

dariotarantini:
I propose to also add opcodes for modular aritmethics, like:

EmelyanenkoK:
To further simplify long arithmetic operations it is suggested to define `d=0` mode in `A9mscdf` op-code the following way:
it is equivalent to `d=3`, however prior to division or shift additional integer number `w` from the stack is added to the intermediate number.
For instance 
- `MULDIV` `(xy)/z` transforms to `(xy+w)/z` - `MULADDDIV`
- `MULDIVMOD` transforms to `MULADDDIVMOD` - `(x y w z -- q r), q = floor((xy+w)/z), r = xy+w-zq`

oleganza:
Strongly recommend adding anything EC-related as Ristretto API with Merlin: https://ristretto.group / https://merlin.cool

oleganza:
Would be nice to have perfect encapsulation for recursive execution of continuations. This would allow us to cheaply execute untrusted code in order to get simple "yes/no" or numeric result while providing asset safety guarantees to the users. 

E.g. we should bring `RUNVM` (or something like that) to TVM that remembers the entire state of the VM on a separate stack, restores them after execution and then returns the nested VM results for introspection in the outer program.

Use-case: predictable and safe standardized Jetton code that allows embedding policy as executable code. That code has static (known AOT) gas limits and can only reply "allow/deny" — that is, cannot mess up accounting.

ex3ndr:
Yes, having something like "RUNVM" that would isolate the contract and its stack is very, very important since, in the TACT compiler, we are executing "init" functions that could theoretically become a vulnerability.

ex3ndr:
> Strongly recommend adding anything EC-related as Ristretto API with Merlin:
> https://ristretto.group/
> https://merlin.cool/
> Ristretto uses the same Curve25519 and its privkeys can be mapped to compatible Ed25519/X25519 pubkeys.

Please, can we just avoid esoteric zkp in ton. There are a lot of tools for ETH and we don't need to slow down innovation here.

oleganza:
> Please, can we just avoid esoteric zkp in ton. There are a lot of tools for ETH and we don't need to slow down innovation here.

I disagree on that take. Ristretto is the opposite of "esoteric zkp". It is a rectification of the original Curve25519 design.

Let me clarify.

Curve25519 is a designed for speed, but does not give prime-order group out of the box. Since it was needed for only two protocols at the time — Schnorr signatures and Diffie-Hellman key exchange, — its author used a couple of hacks to make those specific protocols work. One of them is crude multiplication by 8 that kills the cofactor and eliminates specific risks of small subgroup attacks.

Due to these hacks, a whole lot of things relevant to blockchains are ±broken with Ed25519/X25519 protocols:

* Deterministic key derivation (BIP32) is not possible due to non-linear bit-twiddling in Ed25519.
* Batch signature verification is inconsistently defined.
* DH scheme uses different format for the keys than the Schnorr scheme.
* Performing ElGamal encryption and Schnorr proofs on it can't be cleanly composed.
* Advanced ZKP like Bulletproofs won't work.

What is Ristretto?

* It is a mathematically clean prime-order group.
* It is a pair of encoding/decoding functions over Curve25519 that uses the same point format as in Ed25519. 
* All the performance gains and under-the-hood implementations of Curve25519 remain.
* You can use the same key pair cleanly for signatures, D-H, ElGamal, and key derivation.
* Things compose without hacks.
* You can always convert Ristretto keypair to the existing Ed25519 keys for compatibility.

EmelyanenkoK:
To give feedback on suggestions:
as it is now, we plan to 
1) discard "hashers" in favor of hash operations which pull data elements from the stack
2) global_id and bounce-on-action-fail opcodes
3) add  *ADDDIV* operations 
4) runvm opcode with signature close to 'runvmx' in fift

Discussion about special curve operation (except signature checks) are currently postponed until we understand how cheap those operations can be on current tvm.

ex3ndr:
> > Please, can we just avoid esoteric zkp in ton. There are a lot of tools for ETH and we don't need to slow down innovation here.
> 
> I disagree on that take. Ristretto is the opposite of "esoteric zkp". It is a rectification of the original Curve25519 design.
> 
> Let me clarify.
> 
> Curve25519 is designed for speed, but does not give prime-order group out of the box. Since it was needed for only two protocols at the time — Schnorr signatures and Diffie-Hellman key exchange, — its author used a couple of hacks to make those specific protocols work. One of them is crude multiplication by 8 that kills the cofactor and eliminates specific risks of small subgroup attacks.
> 
> Due to these hacks, a whole lot of things relevant to blockchains are ±broken with Ed25519/X25519 protocols:
> 
> * Deterministic key derivation (BIP32) is not possible due to non-linear bit-twiddling in Ed25519.
> * Batch signature verification is inconsistently defined.
> * DH scheme uses different format for the keys than the Schnorr scheme.
> * Performing ElGamal encryption and Schnorr proofs on it can't be cleanly composed.
> * Advanced ZKP like Bulletproofs won't work.
> 
> What is Ristretto?
> 
> * It is a mathematically clean prime-order group.
> * It is a pair of short encoding/decoding functions (20 LOC each) over Curve25519 that uses the same point format as in Ed25519 under-the-hood.
> * All the performance gains and under-the-hood implementations of Curve25519 remain.
> * You can use the same key pair cleanly for signatures, D-H, ElGamal, and key derivation.
> * Things compose without hacks.
> * You can always convert Ristretto keypair to the existing Ed25519 keys for compatibility.
> 
> Compared to the the rest of the ZKP scene, Curve25519 uses conservative assumptions (no pairings, pure discrete log hardness), its design is battle-tested for 11 years and has been reviewed by possibly more eyes than all other ZKP projects' combined.

Almost nothing makes sense in this message. Why use curve25591 if we need to integrate with existing tools and research of ZKP? There are no reason not to support other curves.

alfredonodo:
zkp and zk-snark are very important in order to be able to implement [private transactions](https://github.com/ton-society/ton-footsteps/issues/22) and thus private workchains.

ex3ndr:
There is a need for commiting `c4` (contract data) without commiting `c5` and bouncing a message in the case of failure after commit. This is required for lazy deployment to make it transparent for users.

ex3ndr:
1) Isn't sha256 already exist? Whats the difference?
2) Since we are adding more curves, is it possible to add any classical zkp one?

EmelyanenkoK:
1. Current sha256 is limited to 127 bytes. New opcodes allows hashing of arbitrary long bytestrings consumed from stack as slices
2. We consider BLS12-381 as most common now (except maybe bn254?). BLS12-381 is more secure and performant than bn254 while can be used the same way

ilyar:
In the recent [TVM Upgrade (#686)](https://github.com/ton-blockchain/ton/commit/9f93888cf402f8421fef38406b67886be043ac58) a new useful opcode was added:
```fift
x{F802} @Defop GASCONSUMED
// ( -- g_c) returns gas consumed by VM so far 26 gas
```
which actually replaces the opcode:
```fift
x{F802} @Defop BUYGAS
// (x -- ) computes the amount of gas that can be bought for x nanograms,
// and sets g_l accordingly in the same way as SETGASLIMIT
```

Which is still described in the document [tvm.pdf](https://ton.org/tvm.pdf) but has not yet been implemented.

Please reveal the details of the reasoning behind the decision.

In my research, I realized that this code is implemented in an alternative implementation of TVM

I understand that there are no obvious reasons to strive for full compatibility of all possible implementations, especially since the process should be arranged in reverse, given that the reference implementation is here.

At the same time, we have the development of TBM technology in more than two third-party projects, it may be useful to take care of compatibility, thus supporting the development of a new TBM-compatible blockchain database.

I'm interested in learning more about the strategic plans for the development of the virtual machine and the criteria on the basis of which decisions are made.

I also think there should be one source of truth, for example [tvm.pdf](https://ton.org/tvm.pdf). in the case of the `GASCONSUMED` opcode, it should not take the place of `BUYGAS` because the latter took it earlier in the specification according to which other implementations can be implemented.

ilyar:
https://github.com/ton-blockchain/ton/pull/768

EmelyanenkoK:
`GASCONSUMED` and unimplemented yet `BUYGAS` serves different purpose.
`GASCONSUMED` allows you to understand how much gas was already used. It doesn't change gas limits or anything, just adds integer to stack

`BUYGAS` increase gas limits by amount of gas equivalent to some amount in TONs.

These two instructions do not replace each other (well they can even help each other actually). However, in contrast to `GASCONSUMED` which can not be implemented in current TVM, `BUYGAS` can be emulated by slightly more expensive consequences of other existing operations, thus implementing of `BUYGAS` was not a high priority.

ilyar:
@EmelyanenkoK  Thanks for the detailed explanation.
There is also a thinning why the `GASCONSUMED`  opcode uses code `x{F802}`? This code is described [in the specification as a BUYGAS opcode](https://github.com/ton-blockchain/ton/blob/6e5145305694debad7a95a838cb7e6b4bba0f7e3/doc/tvm.tex#L2201), it would be logical for a new opcode to use a new code, for example `x{F807}` from reserved for gas-related primitives.

EmelyanenkoK:
@ilyar `GASCONSUMED` was changed to `x{F807}` for compatibility reasons: https://github.com/ton-blockchain/ton/commit/030ebaf772bf95807ccc07520018f9477c241414 
Thank you.

text/0088-tvm-update.md
4304417: 4304417

SalmanAhmrd: UQAP7O1Pm6kHiSerD6L52UIXPYyeYId2ha5WsHWPwOCx_-3p

MrG9090 (approved):

MrG9090 (approved):

PR #89, opened 2022-09-08
TEP: discoverable jetton wallets

EmelyanenkoK:


sasha1618:
This is a crucial feature for the DEX systems that allows a Pool(Pair) rely on Jetton Minter+Jetton Discovery addresses only and do not store jetton wallet address to the init data.

Also, there could be some solutions where this feature is usefull, e.g. smart contract that is created dynamically and receives payments in jettons.

This proposal solves the missing parts of the current Jetton implementation.

text/0089-jetton-wallet-discovery.md
```diff
@@ -0,0 +1,71 @@
...
+Jettom Master should handle message
+
+`get_wallet_address#418cbb4e query_id:uint64 owner_address:MsgAddress include_address:Bool = InternalMsgBody;`
+
+with TON amount higher than `5000 gas-units + msg_forward_prices.lump_price + msg_forward_prices.cell_price` = 0.0061 TON for current basechain settings (if amount is less than that it is not possible to send response) attached
+
+and either throw an exception if amount of incoming value is not enough to calculate wallet address or
+response with message (sent with mode 64)
+
+`report_wallet_address#ad30f94a query_id:uint64 wallet_address:MsgAddress owner_address:(Maybe ^MsgAddress) = InternalMsgBody;`
```
shaharyakir, line 34: Please consider naming report_wallet_address return_wallet_address, because report has the context of marking an entity as malicious etc.

talkol: other potential terms instead of "report" are "return" / "query" / "get"

EmelyanenkoK: Currently we mimic names used in NFT for `get_static_data`.
I don't like return/query naming because I think it will looks a little bit counterintuitive to implement "return_wallet_address" method on recipient side to actually handle the response.
What about a `take_static_data`? Alternatively we can just use `get_static_data` op-code with high-bit set to true (to emphasize it is response), however we doesn't have TL-B notation for that.
Anyway, I don't think that this issue is important and any decision will work ok.

sasha1618: As I can see, the best way to solve the problem is to use a synonym of the word "report". 

I suggest using the word "inform". It is neutral and doesn't have a negative meaning. (e.g. inform_wallet_address, inform_static_data)

PS: imho, we should decide faster so people can start to use an up-to-date jetton standard.

sasha1618: We realised that op should represent the **operation that receiver should execute**.

With that in mind the best way to name our ops is: provide_wallet_address and take_wallet_address.

- Smart-contract sends op:provide_wallet_address.
- Discoverable minter receives op:provide_wallet_address and sends op:take_wallet_address.
- Smart-contract receives op:take_wallet_address and handle the message.

EmelyanenkoK:
@NickNekilov @ex3ndr As independent but with experience and interest in the field experts, please review this TEP

NickNekilov:
@EmelyanenkoK For me, it looks good.
It solves the existing problem of wallet discovery.
Great job!

hacker-volodya:
This pull request enters its Final Comment Period. It ends in 10 calendar days, 03.10.2022.

Gusarich:
LGTM!

hacker-volodya:
FCP is over.

freakyy2 (approved):

MrG9090 (approved):

MrG9090 (approved):

kkomiBot (approved):

text/0089-jetton-wallet-discovery.md
```diff
@@ -9,45 +9,45 @@
 
 # Summary
 
-This proposal suggest to extend standard Jetton master by adding mandatory onchain `get_wallet_address` handler.
```
dommexx20: ZnrYJ1b4uqde3Tw9kJLEsh11H3Z1AQLCGCv3qGcXotM

PR #93, opened 2022-09-12
TEP: Safe Signing

ex3ndr:
Proposal for a Safe Signing scheme.

### DIFF
--- a/text/0093-safe-signing.md
+++ b/text/0093-safe-signing.md
@@ -1,0 +1,73 @@
+- **TEP**: [0](https://github.com/ton-blockchain/TEPs/pull/0) *(don't change)*
+- **title**: Safe Signing
+- **status**: Draft
+- **type**: Core
+- **authors**: [Steve Korshakov](https://github.com/ex3ndr)
+- **created**: 12.09.2022
+- **replaces**: [TEP-0](https://github.com/ton-blockchain/TEPs/blob/master/0000-template.md)
+- **replaced by**: -
+
+# Summary
+
+Safe signing for non-transaction data.
+
+# Motivation
+
+To prove ownership of an address for off-chain services, it is required to provide signing of arbitrary data from wallets. Unfortunately, since key is the same we have to construct a safe way to sign to avoid malicilous signing of a transaction. This specifications does exactly that.
+
+# Guide
+
+To sign a transaction (or just a cell) for TON it is required to create **Cell Representation**, hash it and then sign. Our goal is to create prefix that would never lead for correct
+representation.
+
+From whitepaper:
+
+```
+3.1.4. Standard cell representation. 
+When a cell needs to be transferred by a network protocol or stored in a disk file, it must be serialized. 
+The standard representation CellRepr(c) = CellRepr∞(c) of a cell c as an octet (byte) sequence is constructed as follows:
+1. Two descriptor bytes d1 and d2 are serialized first. Byte d1 equals r+8s+32l, where 0 ≤ r ≤ 4 is the quantity of cell references
+contained in the cell, 0 ≤ l ≤ 3 is the level of the cell, and 0 ≤ s ≤ 1 is 1 for exotic cells and 0 for ordinary cells. Byte d2 equals 
+⌊b/8⌋+⌈b/8⌉, where 0 ≤ b ≤ 1023 is the quantity of data bits in c.
+2. Then the data bits are serialized as ⌈b/8⌉ 8-bit octets (bytes). If b is not a multiple of eight, a binary 1 and up to six 
+binary 0s are appended to the data bits. After that, the data is split into ⌈b/8⌉ eight-bit groups, and each group is interpreted as 
+an unsigned big-endian integer 0 . . . 255 and stored into an octet.
+3. Finally, each of the r cell references is represented by 32 bytes contain- ing the 256-bit representation hash Hash(ci), explained
+below in 3.1.5, of the cell ci referred to.
+In this way, 2 + ⌈b/8⌉ + 32r bytes of CellRepr(c) are obtained.
+```
+
+From documentation it is obvious that if we want to sign cell with a single reference it will have a form if `<prefix><referehce_hash>`. Now we need to construct prefix the way representation became invalid. It is obvious that values `5 ≤ r ≤ 7`, `2 ≤ s ≤ 3` and `4 ≤ l ≤ 7` are all invalid. We will pick maximum values and this would be `0xff` byte. While this already makes representation invalid we will strength it even further just in case if cell representation specification will be expanded. Second byte doesn't have invalid values so let's just pick the same `0xff` byte. This value would represent `1023` bits, to make it invalid we will simply write to buffer anything that is less than 127 bytes long. This standart defines this as binary representation of string `ton-safe-sign-magic`.
+
+# Specification
+
+To sign an arbitrary cell that couldn't later used as a transaction in blockchain we compute data for signing next way:
+
+```js
+let cell: Cell; // Incoming cell
+let hash = cell.hash(); // Hash of cell
+let data = sha256(Buffer.concat([Buffer.from([0xff, 0xff]), Buffer.from('ton-safe-sign-magic'), hash])); // Data to hash
+let signature = sign(data, publicKey); // Resulting signature
+```
+
+For specific needs it is possible to change `ton-safe-sign-magic` for separating different signature usage, for example for signing "child" keys that could be used as proof of ownership off-chain or on-chain, but this string must be non-zero and not too big to form a 127 bytes string for signing.
+
+# Drawbacks
+
+None
+
+# Rationale and alternatives
+
+None
+
+# Prior art
+
+This poposal is a variation of [Ethereum one](https://eips.ethereum.org/EIPS/eip-1271).
+
+# Unresolved questions
+
+None
+
+# Future possibilities
+
+None


talkol:
I think this is a critical standard that resolves a very important need.

Is it possible to add to the standard an example in FunC that checks the signature to demonstrate that a TVM contract will be later able to verify the offline signature? We need this to implement gas-efficient multi-sigs where one of the parties collects offline signatures from the others and sends a single transaction to the multi-sig contract with all signatures included.

ex3ndr:
For multisig, you can use a default signing.

talkol:
> For multisig, you can use a default signing.

From a dapp developer standpoint I'm mostly interested in the dapp-wallet interface like this one - https://github.com/ton-foundation/ton-x#request-signature

I assume this uses the safe method, right?

I don't think wallets are going to give dapp developers both options (safe and unsafe), so I'm preparing myself to use the safe option always and still want to be able to verify it in smart contracts if needed

ex3ndr:
I see, right now multisig stores all requests onchain thought. 

I think this signature wouldn't be a good fit for signing since it doesn't expect to sign a transaction (that MUST be presented to a user and confirmed).

I am not sure what would be a good way to make this for wallets.

talkol:
Let's ignore the multi-sig example because it's confusing since multi-sig is used for transactions.

Can we make sure that this safe offline signing over a cell is verifiable from TVM?

ex3ndr:
I actually reviewed the idea of verification via TVM and I think this standard is designed to be NOT verifiable from TVM and should remain the same. 

A different signature scheme should be created to make it work in smart contracts, but I am not aware of how to do so yet since there are too many different ways to sign stuff in TON unlike in, for example, ETH where you can distinguish clearly two schemes that are used for non-transactional and transactional signing. Unfortunately, in TON any cell-based signing could be used somewhere to facilitate transactions.

oleganza:
For this to be safe to use across apps and contracts I think we should add mandatory binding for (1) domain name or TON address; (2) timestamp.

(1): well-behaved apps/contracts would check that the message is signed for them, while the wallet would be able to tell the user who would be the consumer of the signed message.

(2): timestamp would be enforced by the wallet to make sure the message is pinned into correct point of time matching the user's intent. Well-behaved apps would add their TTL policy on top of that and reject stale messages and replay attempts.

ex3ndr:
It is much more than that and out of scope of this proposal. We can create a separate standard on top of this. We are using address proofs and it have time stamps, random seeds, etc. but this is out of scope of this standard.

ex3ndr:
It seems that this signing would work in contract:

```func
cell c; ;; Source cell
var hash = cell_hash(c);
var b = begin_cell()
                 .store_uint(0xff, 8)
                 .store_uint(0xff, 8)
                 .store_uint(2596593398242688430396682338969803153947978083, 152)
                 .store_uint(hash, 256)
                 .end_cell();
var target_hash = string_hash(b);
var is_valid = check_signature(target_hash, sig, pk);
```

EmelyanenkoK:
I find this standard well though and ready to be accepted. 
Earlier, I put off discussing the proposal because it was born out of the needs of a single ecosystem player (and the standard should unify solutions for multiple parties).
If now other companies have similar needs (for example, ton-connect [oleganza](https://github.com/oleganza)), let's discuss and accept.

Regarding the issue raised by [oleganza](https://github.com/oleganza). Indeed we surely want an analog of [EIP-712](https://eips.ethereum.org/EIPS/eip-712) for informed signing of data structures, and I agree with [ex3ndr](https://github.com/ex3ndr) that this task is important and complex enough to be an isolated standard. From the other hand, we obviously do not want introduce "blind signing" along with "informed signing", since the former will undermine the last. Are we ready to formulate our analog of EIP-712 to accept both standards simultaneously and keep only "informed signing"?

oleganza:
@ex3ndr could you please add your FunC sample code into the TEP under the "Specification" section? That would nicely show both the offchain (JS) and onchain (FunC) usage.

talkol:
@EmelyanenkoK who will formulate  the analog of EIP-712?

We depend on off-chain for a DAO tools product we’re building and we’ll be happy to PR support into the wallets, but we need the standard to be accepted for that

oleganza:
_Re: EIP-712 and TON Connect._

I think we need to finalize and merge TEP93 as a "base protocol". And for the actual app-facing protocol I suggest creating a separate TEP with the following points:

1. Start with **ton-x** implementation of [requestSign](https://github.com/ton-foundation/ton-x#request-signature) with dual-cell design (user-visible + technical attributes).
2. Sort out whether we use expiration time (current impl in ton-x) or user's current time. TTL is chosen by the app developer either way based on their security model. Not sure if the wallet must do any sanity checks on TTL and impose limits. 
3. Leave the space to be able to tie the message to a domain (TON.DNS, raw pubkey, contract address etc.). Current implementation in ton-x is a free-form user-visible text, which is fine for many serverless dapps. However, even there tying the signature to a contract address (e.g. DEX router) would be a good idea. 

As of all the other structured data, i'd leave it for future expansion based on patterns that would emerge over time. There's a limit on how much the user would validate with their own eyes anyway. And if you want to cross-validate specific messages for specific apps, then there could be custom support for those cases without necessarily a universal fit-all standard.

talkol:
@oleganza I guess we can finalize TEP93 quickly and the extension TEP will take time. Once we finalize TEP93, are you willing to add support for it in TonConnect2 and TonKeeper without waiting for the extension TEP?

oleganza:
@talkol i sure would like to add the support, but ton-x's requestSign is not just TEP93, it is "TEP93 + specific placement of user/machine-readable data" which only exists in one specific implementation. 

I suggest we write it out in one short markdown, check those two points regarding TTL/timestamp and potential domain separation and roll with it.

KuznetsovNikita:
I'm ok with the proposal.

OpenMask already implemented this scheme:

The implementation
https://github.com/OpenProduct/openmask-extension/blob/d9d1a29730357794895ff2cfd1689bec2a4d1dea/src/view/screen/notifications/sign/api.ts#L21

The documetation:
https://www.openmask.app/docs/api-reference/rpc-api#ton_personalsign

The e2e implementation: 
https://www.openmask.app/test-dapp#sign-data


Should we agree about method name and parameters?
In the OpenMask  it's implemented in this way:
```
const signature = await provider.send("ton_personalSign", {
    data: "Hello world",
});
```

talkol:
@oleganza that sounds great, can you outline the markdown you're comfortable with for the initial implementation? I'll chase @ex3ndr to get his support as well so we can agree on the first milestone that will add off-chain to wallets

talkol:
Once we do that, my team can help PR TobHub, TonKeeper, OpenMask and others so all are compatible and produce a similar format so anyone trying to verify won't lose their mind

mr-tron:
>     2. Sort out whether we use expiration time (current impl in ton-x) or user's current time. TTL is chosen by the app developer either way based on their security model. Not sure if the wallet must do any sanity checks on TTL and impose limits.

It depends on does user allow to change TTL. If wallet doesn't have gui for changing TTL, where is no sense to use TTL instead of current timestamp.

>     3. Leave the space to be able to tie the message to a domain (TON.DNS, raw pubkey, contract address etc.). Current implementation in ton-x is a free-form user-visible text, which is fine for many serverless dapps. However, even there tying the signature to a contract address (e.g. DEX router) would be a good idea.
Totally agree what using TON DNS as part (optional) of common signature scheme is a good idea. If it's visible for user and marked as domain separated from plain text.

KuznetsovNikita:
I think we should agree the wallet interface.

I like the idea to have a user-visible + technical attributes messages.
Something like this:
```
const signature = await provider.send("ton_personalSign", {
    data: "Some random string",
    text: "Please, sign our terms or service and privacy policy"
});
```

For my perspective it would be enough. 

The `signature` may be store in the contracts and then verify there.
In case if the dApp developer would like to add an expiration data, he may just store `timestamp` next to `signature`.

siandreev:
If we want to make the convention extensible, then we need to add a lot of data to the target cell.
We need to add a schema version.
We also need to add a timestamp. It is up to the verifier of the signature to decide whether it has expired for him or not. 
In addition, we need to add specific ref cells depending on the schema version.
So, when using the way proposed by @ex3ndr as is, a very large nesting of cells is obtained.

```
signature(hash(0xffff ++ prefix ++ hash(X)))

X = Cell {
  bits: version ++ timestamp,
  refs: [Cell{text}, Cell{payload}]
}
```


I think we can replace the `store_uint(2596593398242688430396682338969803153947978083, 152)` prefix with fixed-length `schema_version` ++ `timestamp`.
Then the target cell X will contain data depending on the schema version.

```
signature(hash(0xffff ++ uint32 schema_version ++ uint64 timestamp_unix_seconds ++ hash(X)))

X_version0 = Cell {
  bits: <utf8_text>,
  refs: [Cell{payload}]
}

X_version1 = Cell {
  bits: <contract_address>,
  refs: [Cell{payload}]
}

X_version2 = Cell {
  bits: <ton.dns name>,
  refs: [Cell{payload}]
}
```

So, the current ton-x implementation of the request_sign will almost match the "version 0" schema.

Updated func code looks like
```
cell x; ;; Source cell
var hash = cell_hash(x);
var b = begin_cell()
                 .store_uint(0xff, 8)
                 .store_uint(0xff, 8)
                 .store_uint(schema_version, 32)
                 .store_uint(timestamp_unix_seconds, 64)
                 .store_uint(hash, 256)
                 .end_cell();
var target_hash = string_hash(b);
var is_valid = check_signature(target_hash, sig, pk);
```

KuznetsovNikita:
> ```
> X_version0 = Cell {
>   bits: <utf8_text>,
>   refs: [Cell{payload}]
> }
> 
> X_version1 = Cell {
>   bits: <contract_address>,
>   refs: [Cell{payload}]
> }
> 
> X_version2 = Cell {
>   bits: <ton.dns name>,
>   refs: [Cell{payload}]
> }
> ```

@siandreev To verify signatures on the contract level developers have to deliver or store a lot's of information

- wallet public key 
- timestamp
- signature
- also developers have to store hash for payload or payload components:
```
var X_version0 = begin_cell()
                 .store_slice("text")
                 .store_ref(begin_cell().store_slice("payload").end_cell())
                 .end_cell();
 ```
 
 ```
var X_version1 = begin_cell()
                 .store_slice(my_address())
                 .store_ref(begin_cell().store_slice("payload").end_cell())
                 .end_cell();
 ```
 
  ```
var X_version1 = begin_cell()
                 .store_slice("wallet.ton")
                 .store_ref(begin_cell().store_slice("payload").end_cell())
                 .end_cell();
 ```

In general all of this target cell X almost the same
1) Is it really necessary to split them into versions?

**Comment by KuznetsovNikita** [comment]
> ```
> cell x; ;; Source cell
> var hash = cell_hash(x);
> var b = begin_cell()
>                  .store_uint(0xff, 8)
>                  .store_uint(0xff, 8)
>                  .store_uint(schema_version, 32)
>                  .store_uint(timestamp_unix_seconds, 64)
>                  .store_uint(hash, 256)
>                  .end_cell();
> var target_hash = string_hash(b);
> var is_valid = check_signature(target_hash, sig, pk);
> ```

@siandreev 
To verify signature the developer have to store `timestamp_unix_seconds` in a contract. 
Storing timestamp will require for all contracts.

For my perspective we may make it optional. Contract developers may include timestamp to  target cell X in case if there would like to create a relation with timestamp.
```
var X_version0 = begin_cell()
                 .store_slice("text")
                 .store_uint(timestamp_unix_seconds, 64)
                 .store_ref(begin_cell().store_slice("payload").end_cell())
                 .end_cell();
```

Is it better to make storing timestamp required?

**Comment by oleganza** [comment]
@KuznetsovNikita not sure I understood your concern. Let me clarify. This is what would be the data sent into the contract where X stands for the payload cell:

```
Cell {
  bits: signature (512 bits) ++ version (32 bits) ++ timestamp (64 bits)
  refs: [ X ]
}
```

This assumes that the user's pubkey is already stored in the contract or transmitted somehow.

The contract is supposed to read the signature, hash the rest of the slice and verify the signature. Then it may proceed checking app-specific policies: TTL against the timestamp, verify the version and the contents of the payload X. 

Does not seem to me that timestamp adds significant overhead, and it is necessary to make sure signatures are pinned in space _and time_.

oleganza:
Yesterday night I and @mr-tron have discovered that the trick with invalid cell representation is unnecessary. 

The documentation on ton.org incorrectly states that `check_data_signature(slice,...)` adds another round of sha256 hashing. This is not the case: the variable-length binary string is passed as-is to Ed25519 scheme ([crypto/vm/tonops.cpp:367](https://github.com/ton-blockchain/ton/blob/master/crypto/vm/tonops.cpp#L367-L381)), just like the variant that accepts uint256. This allows us to disambiguate data signatures and transactions signatures simple by making the message size different from the 256 bits (as used by the wallet contracts).

In the alternative proposal (https://github.com/ton-blockchain/TEPs/pull/104) that string is a 352-bit string with its own domain separator ("schema version"), timestamp and a hash of a payload cell.

PR #95, opened 2022-09-16
TEP: Proof of NFT Ownership

xssnick:
Onchain ownership proofs for NFT, standard extension.


text/0095-prove-ownership.md
```diff
@@ -0,0 +1,239 @@
...
+```
+request_owner#d0c3bfea query_id:uint64 dest:MsgAddress 
+forward_payload:^Cell with_content:Bool = InternalMsgBody;
+```
+After that NFT will send transfer to `dest` with scheme:
+```
+owner_info#c2fa9387 query_id:uint64 item_id:uint256 initiator:MsgAddress owner:MsgAddress 
+data:^Cell content:(Maybe ^Cell) = InternalMsgBody;
+```
+If something goes wrong and target contract not accepts message, and it will be bounced back to NFT, NFT will proxy this bounce to **initiator**.
```
oleganza, line 146:
```suggestion
If the target contract fails to process the message, it will be bounced back to NFT, NFT then should proxy that bounced message to the **initiator**.
```

oleganza, line 146: Should the NFT check that `initiator` != `dest`? Imagine a (supposedly) common case when a contract requests a proof for itself and does not process the request properly. Seems like it will get spammed with a bounce. Would we avoid an infinite back-and-forth in such case?

oleganza:
The spec looks fine for me. A personal cosmetic preference is to have a more "active"-sounding name for ownership proof from the owner that works better with use-cases like voting. Maybe something like "authenticate"? The idea would be that the person would use NFT as an authentication token to send some request wrapped in such authentication envelope. "Prove ownership" sounds like a thing in and of itself, rather than a wrapper for some action of interest.

PR #97, opened 2022-10-06
TEP: Universal wallet

ex3ndr:
This proposal introduces universal wallet that could be upgraded to a latest version by validator voting


text/0097-universal-wallet.md
```diff
@@ -0,0 +1,56 @@
...
+
+The wallet itself have code (Fift):
+```
+PUSHINT 74
+CONFIGPARAM
+BLESS
+JMPX
+```
+
+And inital data is `seq:int64 public_key:int256 wallet_id:int64`.
```
Gusarich, line 34: What if in new version of the wallet there will be more variables in data? How will it upgrade then?

ex3ndr: Code must support the absence of any added field. We are doing this in our nominators pool and it worked well.

EmelyanenkoK:
As it was privately discussed, issues of the standard, in particular transferring too much power over users' wallets to validators, can be alleviated by changing proposal in such a way that user will authorize code update.
In other words wallet code will consist of 
- function which allow to get new-code-library-hash from config parameter and save it as current "executable" code.
- executor of current "executable" code.

This way, from one hand, user can avoid upgrading to unwanted wallet version (and thus validators will be unable for instance sanction some wallets and thus will not be responsible for not doing it), and from the other we will not open attack vector by allowing standard wallet to be upgraded to any new code: only upgrade to code authorized by validators will be possible.

PR #98, opened 2022-10-14
Update 0064 add render type for Jetton

ivklim-ton-play:
The idea is to allow projects to flag how best to display their tokens in wallets, markets, etc.
This small introduction will increase the attractiveness of projects for which visual display is important.
These objects are optional.

For example, in games it is important that the tokens are displayed as a large card and look more like an NFT than a currency. At the same time, the game can also have a currency.


oleganza:
ack

EmelyanenkoK:
Since this update is supported by developers of main explorers and wallets it gets merged.

PR #104, opened 2022-12-14
TEP: Data signatures

oleganza:
This proposal is an evolution of @ex3ndr's [safe-signing proposal](https://github.com/ton-blockchain/TEPs/pull/93).

Highlights:

* "Invalid cell representation" trick is not necessary because  we are signing a differently-sized message.
* Separate schemas for plaintext messages and binary messages. User can explicitly check all the data they are signing.
* Mandatory binding to a timestamp.
* Support for binding to smart contract address and domain name.
* Possibility for future extensions with detailed schemas retaining backwards compatibility with the wallets.

Rendered specification:
* [📄 TEP: Data Signatures](https://github.com/oleganza/TEPs/blob/datasig/text/0000-data-signatures.md)

### DIFF
--- a/text/0000-data-signatures.md
+++ b/text/0000-data-signatures.md
@@ -219,10 +219,6 @@ Applications enforce domain separation to reject signatures from other domains.
 
 Wallets enforce domain separation to protect users from inadvertently signing for another app.
 
-## Why messages are limited to 127 bytes?
-
-Current proposal offers the simplest text encoding using 1023 contiguous bits available in the payload cell. More complex text encoding schemes (e.g. chunked) are possible to add to this protocol later if needed.
-
 ## How to use contract address binding
 
 Wallets may display a name and an icon for the well-known contracts and even interpret parameters as specific actions.
--- a/text/0000-data-signatures.md
+++ b/text/0000-data-signatures.md
@@ -1,0 +1,227 @@
+- **TEP**: [0](https://github.com/ton-blockchain/TEPs/pull/0) *(don't change)*
+- **title**: Data Signatures
+- **status**: Draft
+- **type**: Core
+- **authors**: [Steve Korshakov](https://github.com/ex3ndr), [Oleg Andreev](https://github.com/oleganza), [Sergey Andreev](https://github.com/siandreev)
+- **created**: 13.12.2022
+- **replaces**: [TEP-0](https://github.com/ton-blockchain/TEPs/blob/master/0000-template.md)
+- **replaced by**: -
+
+# Summary
+
+Safe signing for non-transaction data by the wallet’s key.
+
+# Motivation
+
+User’s wallet in TON ecosystem is not only used to perform asset transfers, but also plays a role of a universal identifier for on-chain and off-chain application.
+
+Most commonly, the wallet is used to send TON messages that transfer coins, change ownership of tokens and interact with other smart contracts. In all these cases the client app signs a well-formed transaction that can be verified by the wallet’s public key.
+
+The proposed protocol addresses scenarios where the wallet’s public key verifies arbitrary data used within contracts other than the user’s wallet (that is, non-transaction data) or offchain applications.
+
+# Examples
+
+1. Proving ownership of an address for off-chain services.
+2. Signing a structured message to be verified in a TON smart contract (inside TVM).
+
+# Safe hashing
+
+To sign a transaction (or just a cell) for TON it is required to create **Cell Representation**, hash it and then sign. Our goal is to create a signature scheme that guarantees to never collide with the signed cells, and therefore guarantee that this scheme cannot be abused to send transaction on behalf of the wallet.
+
+From the TON whitepaper:
+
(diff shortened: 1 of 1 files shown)


oleganza:
Received a number of suggestions from @rulon: 

1. Plain text signatures are safer to implement without any hidden binary payload. "What you see is what you get". I agree, will amend the spec to reflect that.
2. Instead of incremental schema versions we could use crc32 of the TL-B description of the payload.
3. We could combine both domain and address together in one schema. In case you have both the service and the target contract to use the data with.

oleganza:
Update the spec based on feedback from @EmelyanenkoK & @mr-tron. 

* Plaintext messages use chunked text encoding;
* Scheme 1+2 are merged: one may use both the contract address and the domain name;
* Explicit specification for the domain name encoding;
* Domain name is also chunk-encoded for consistency and to avoid unnecessary length limits.

oleganza:
Based on feedback by @tolya-yanot replaced chunked encoding with SHA256. This comes at no tradeoff for domain names: they are transmitted or known explicitly anyway. 

As for the plaintext string: it is unlikely that smart contracts would verify complex text strings, and yet it is possible to hash them or introduce a dedicated TL-B scheme for that purpose in the future. For short "command" strings the contracts could verify sha256 images directly against the built-in constants.

oleganza:
One more update: @mr-tron pointed out two things:

1. Cells already provide hashing by themselves, so it would be prudent to use snake format (as in [TEP-64](https://github.com/ton-blockchain/TEPs/blob/master/text/0064-token-data-standard.md)) for text message and domain.
2. It is not necessary to perform the "invalid cell representation" hack if we avoid extra hashing before computing/verifying the signature. We could simply send 352-bit message into Ed25519 as-is and that would provide enough separation from any wallet transaction (that use 256-bit cell hash as a message in the Ed25519 protocol).

oleganza:
A couple of minor tweaks:

1. Domain name is embedded into cell to allow replacing its explicit content with its hash: might be handy in some situations.
2. There was an idea to embed TL-B description of the payload and automatically display it to the user, but this requires more time to design, so we are adding `ext:(Maybe ^Cell)` for such future extension.

EmelyanenkoK:
LGTM

talkol:
LGTM, thanks all for the hard work to finalize this as part of TC2

oleganza:
Do we need to explicitly bind these signatures to the network ID (-239 for mainnet, -3 for testnet) to avoid ambiguity between networks?

Also: in TON Connect there is custom "address proof" scheme that we may want to replace with this scheme for consistency.

mr-tron:
Your wallet is already have subwallet id inside address which should be first 4 bytes of zero state hash.
https://github.com/ton-blockchain/ton/blob/4b940f8bad9c2d3bf44f196f6995963c7cee9cc3/tonlib/tonlib/TonlibClient.cpp#L2420
I don't see any sense to add another network id which everyone will ignore.

ex3ndr:
I am reserving the prefix of `plaintext text:Slice = PayloadCell` == `96890e83` for ledger signatures. It differs in that it just signs a hash of a string instead of a cell, also doesn't have an expiration date, etc. It is challenging to implement time verification anyway.

oleganza:
> Your wallet is already have subwallet id inside address which should be first 4 bytes of zero state hash.
> https://github.com/ton-blockchain/ton/blob/4b940f8bad9c2d3bf44f196f6995963c7cee9cc3/tonlib/tonlib/TonlibClient.cpp#L2420
> I don't see any sense to add another network id which everyone will ignore.

Do you mean using the same 4 bytes of the zero state hash as an identifier? Note that the current proposal does not bind signatures to any wallet address (which is bound to the network id as you described).

oleganza:
> I am reserving the prefix of `plaintext text:Slice = PayloadCell` == `96890e83` for ledger signatures. It differs in that it just signs a hash of a string instead of a cell, also doesn't have an expiration date, etc. It is challenging to implement time verification anyway.

I don't understand the use-case. The snake format effectively compresses any string to a hash (for the tail) and allows .

Anyway, I'd suggest to name this something like "hashsig hash:Slice = PayloadCell" to make the difference clear. But then it's only useful if you don't plan to display the contents of the text to user, which would be weird and insecure.

Note that the timestamp exists outside the payloads. If you want to make an eternal signature you may use all-zero timestamp and if it's important, the clockless HSM may check that the time is all-zero so that you know it's not going to have an expiration time. However, even such HSM may format the time and display it for sanity check to the user.

oleganza:
Re: binding to the network. These signatures are going to be used also offchain, so we can delegate to the dapps to embed additional binding in their custom payloads and not include the network ID in this standard. E.g. if a signed message operates a smart contract on a certain network, then it may be enough to bind through a contract address (that should be different on a testnet).

mr-tron:
I think should add standard prefix to any signed payload which contains ton connect magic prefix (like version maybe), ton connect domain and signing timestamp for avoid reusing signature in other projects

PR #108, opened 2023-01-04
🔧 ABI

ex3ndr:
Draft of an ABI standard

### DIFF
--- a/text/0108-abi.md
+++ b/text/0108-abi.md
@@ -1,0 +1,68 @@
+- **TEP**: [108](https://github.com/ton-blockchain/TEPs/pull/108) _(don't change)_
+- **title**: ABI
+- **status**: Draft
+- **type**: Core
+- **authors**: [Steve Korshakov](https://github.com/ex3ndr) _(replace)_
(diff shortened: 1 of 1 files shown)


TrueCarry:
Getters have type, but what about getters that return tuples like get_nft_data, get_sale_data? Maybe instead of `type` it should be output: [BasicType]?
Also it would be great to have ABI represenation not only for methods/messages, but also for cells. In many cases cell layout is static and can be defined beforehand.

ex3ndr:
Return types just a normal types. Right now it is just a struct in TACT and fields are mapped to tensor. So you define it as a normal struct/message

andreypfau:
I have an extremely negative attitude to the idea of using JSON as an ABI for TON contracts in general. We have excellent ways to sterilize structures for cells - TLB and it is he who should be used as an ABI

ex3ndr:
As always i see TF is fully delusional claming that we have "tlb" while in a YEAR no one was able to build a parser to any language. half of the parsing code in core is a manual parsing and not automated at all.

PR #111, opened 2023-02-07
Create TEP-NFT-Burning.md

dima-brook:
XP.NETWORK Bridge requires burning functionality to properly remove wrapped assets when transferring NFTs / SFTs to other chains from TON.


dima-brook (commented):
Updated the author & date

mr-tron:
Proposal doesn't describe HOW it should be implemented

PR #115, opened 2023-02-27
TON Connect

tolya-yanot:
The Tonkeeper team has done a great job. They studied what's been done in other blockchains, took the best approaches and improved the weaknesses.

If you like comparisons, TON Connect can be seen as an improved and most importantly decentralized version of WalletConnect.

Many TON dapps and wallets already support TON Connect. With the upcoming updates of the TON Connect SDK and UI, it's going to be pretty good. Let's have a final discussion about the protocol to make it a network-wide standard.

If you see any flaws or omissions of the protocol, please describe them here.

### DIFF
--- a/text/0000-ton-connect.md
+++ b/text/0000-ton-connect.md
@@ -1,0 +1,121 @@
+- **TEP**: [0](https://github.com/ton-blockchain/TEPs/pull/0)
+- **title**: TON Connect
+- **status**: Active
+- **type**: Core
+- **authors**: @oleganza, @siandreev, subden, brainpicture, @sorokin0andrey, abogoslavskiy, @MariaBit, Olga May, Aleksei Mazelyuk, tonrostislav, KuznetsovNikita.
+- **created**: 20.10.2022
+- **replaces**: -
+- **replaced by**: -
+
+# Summary
+
+TON Connect is a unified protocol for communication between TON wallets and TON apps.
+
+# Motivation
+
+The Open Network needs a **unified** protocol for communication between TON wallets and TON (d)apps to achieve following goals:
+
+1. Any TON app (web / desktop / mobile / etc) can be operated by any wallet (mobile / desktop / web / browser extension / dapp browser / hardware / etc).
+
+
+2. Users get a familiar and friendly experience across all TON apps.
+
+# Guide
+
+Apps provide the UI to an infinite range of functionality in TON based on smart contracts, but do not have immediate access to users’ funds. Therefore they are often called decentralized apps or “dapps”.
+
+Wallets provide the UI to approving transactions and hold users’ cryptographic keys securely on their personal devices.
+
+This separation of concerns enables rapid innovation and high level of security for the users: wallets do not need to build walled-garden ecosystems themselves, while the apps do not need to take the risk holding end users’ accounts.
+
+TON Connect is a bridge that crosses this conceptual gap.
+
+TON Connect, in addition to the transport layer of communication between the wallet and the app, provides methods of authorization, sending transactions to the network, interaction with smart contracts, etc.
+
+# Specification
+
+Docs and specs are in [https://github.com/ton-blockchain/ton-connect](https://github.com/ton-blockchain/ton-connect) repo.
+
+At the time this PR was created (27 Feb, 2023) the revision was `acc5dd4d2106891cbfcade8d7faa58b9e16937fd`.
+
+# Rationale and alternatives
+
+## Unified way
+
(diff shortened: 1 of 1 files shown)


text/0000-ton-connect.md
```diff
@@ -0,0 +1,121 @@
...
+
+Apps provide the UI to an infinite range of functionality in TON based on smart contracts, but do not have immediate access to users’ funds. Therefore they are often called decentralized apps or “dapps”.
+
+Wallets provide the UI to approving transactions and hold users’ cryptographic keys securely on their personal devices.
+
+This separation of concerns enables rapid innovation and high level of security for the users: wallets do not need to build walled-garden ecosystems themselves, while the apps do not need to take the risk holding end users’ accounts.
+
+TON Connect is a bridge that crosses this conceptual gap.
+
+TON Connect, in addition to the transport layer of communication between the wallet and the app, provides methods of authorization, sending transactions to the network, interaction with smart contracts, etc.
```
AndreyDodonov-EH, line 33: Which methods of authorization are implied here?
Usually local authorization is done in wallet itself (fingerprint, pin, password).

Is it supposed to be something other than just private key (or complete seed phrase) + local authorization?

E.g. some types of 2FA to sacrifice some decentralization in favor of security?

Gusarich: Authorization in this context means proving to the app backed that user really owns the wallet.
It is also shown in docs: https://ton.org/docs/develop/dapps/ton-connect/integration#logging-out-and-requesting-tonproof

AndreyDodonov-EH: Thanks a lot!
Need to go through this thoroughly.

stels-cs:
Add way to abort transaction sent via ```connector.sendTransaction```

like abortable fetch in browser api https://developer.chrome.com/blog/abortable-fetch/ its use AbortSignal https://developer.mozilla.org/ru/docs/Web/API/AbortSignal

If the user changes his mind about sending a transaction, he should be able to exit of the UI. Like on screen.
It's may happen after call ```connector.sendTransaction```, in this case dapp can hide own UI, but can't abort transaction in wallet UI
 image

Of course, this does not affect the blockchain, only UI

tvorogme:
Currently, `connector.sendTransaction` returned signed boc. For better API calls, I suggest to return transaction hash (wallet is 100% sure that the transaction is processed, so it must have a hash of it?). It's quite hard to find a transaction by signed boc on scale of 100k+ users.

tolya-yanot:
> Currently, `connector.sendTransaction` returned signed boc. For better API calls, I suggest to return transaction hash (wallet is 100% sure that the transaction is processed, so it must have a hash of it?). It's quite hard to find a transaction by signed boc on scale of 100k+ users.

Maybe it can be done as a separate event - immediately at the time of sending, return a message or message hash, and then send an event with a hash of the transaction

tolya-yanot:
Since some wallets (hardware wallets or custom wallets such as a lockup wallet) will probably never support multiple messages in one SendTransaction, the parameter `maxMessages: number` added to the Feature `SendTransaction`.

The `Feature` format has also been changed so that the wallet can return more information if required.

before:

`type Feature = 'SendTransaction' | 'SignData';`

after:

`type Feature = { name: 'SendTransaction', maxMessages: number } | { name: 'SignData' };`

For some time in the process of transition, wallets can return both format `features = [ 'SendTransaction', 'SignData', { name: 'SendTransaction', maxMessages: number } | { name: 'SignData' } ]` for backward compatibility.

Shi8ho11 (approved):

Shi8ho11 (commented):
🙂

tolya-yanot:
> Add way to abort transaction sent via `connector.sendTransaction`
> 
> like abortable fetch in browser api https://developer.chrome.com/blog/abortable-fetch/ its use AbortSignal https://developer.mozilla.org/ru/docs/Web/API/AbortSignal
> 
> If the user changes his mind about sending a transaction, he should be able to exit of the UI. Like on screen. It's may happen after call `connector.sendTransaction`, in this case dapp can hide own UI, but can't abort transaction in wallet UI image
> 
> Of course, this does not affect the blockchain, only UI

NOTE: the response to the request to cancel `sendTransaction` must contain information whether the transaction signing was cancelled or whether the user has already sent it.

tolya-yanot:
I think that it would be nice to implement the proposals https://github.com/ton-blockchain/TEPs/pull/115#issuecomment-1450096183 and https://github.com/ton-blockchain/TEPs/pull/115#issuecomment-1458174944 in one of the next versions of TON Connect protocol.

MrG9090 (approved):

PR #127, opened 2023-09-23
Update TEP-64 with the TON Storage method for storing token data

Gusarich:
As The Open Network continues its growth in Web3 direction, the relevance of TON Storage is intensifying. It has become imperative to standardize the method of storing token metadata within TON Storage.

This Pull Request introduces a new approach to metadata storage using TON Storage. Additionally, it enhances the existing Semi-chain content layout by offering the ability to store supplementary data using TON Storage.

### DIFF
--- a/text/0064-token-data-standard.md
+++ b/text/0064-token-data-standard.md
@@ -116,27 +116,30 @@ snake#00 data:(SnakeData ~n) = ContentData;
 chunks#01 data:ChunkedData = ContentData;
 onchain#00 data:(HashMapE 256 ^ContentData) = FullContent;
 offchain#01 uri:Text = FullContent;
+tonstorage#02 bagid:uint256 = FullContent;
 ```
 
 Note, that while TL-B scheme does not constrain bit size of each chunk it is expected that all chunks contain ceil number of bytes.
 
 ## NFT metadata attributes
 1. `uri` - Optional. Used by "Semi-chain content layout". ASCII string. A URI pointing to JSON document with metadata.
-2. `name` - Optional. UTF8 string. Identifies the asset.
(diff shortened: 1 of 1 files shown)


howardpen9:
nice

tvorogme:
Hey, let's add:

- website
- social

Links to on-chain data keys? Those tags usually used by collections & jetton owners.

EmelyanenkoK:
Do we need additional `path` parameter in `tonstorage#02` for bags with more than 1 file?

mr-tron:
why binary encoding is better than simple OffChain url "tonstorage://BAGBAGID/token.json"? 
Url is already supported by different services like tonkeeper/tonapi/tonviewer and maybe some others.

Gusarich:
> Hey, let's add:
> 
> * "website"
> * "social"
> 
> Keys to on-chain data? Those tags usually used by collections & jetton owners. Both keys must contain string value of URL
> 
> Or we can think about dict where key will be index of link, and value link. To allow several links per 1 item

Hey @tvorogme!
It may be useful to expand the standard for metadata attributes, but I think it's out of scope of this PR.

Gusarich:
> why binary encoding is better than simple OffChain url "tonstorage://BAGBAGID/token.json"?
> Url is already supported by different services like tonkeeper/tonapi/tonviewer and maybe some others.

Well, maybe it's enough. But anyway, we need to standardise it in the TEP.

mr-tron:
Ok. You can write in standard what meta url should have schema from list http|https|tonstorage|ipfs|ipns|ftp|ohshit

Danielflord (approved):

Amnat-auyana1998 (commented):
text/0064-token-data-standard.md

PR #129, opened 2023-12-10
Extension for NFT standart: destroyable NFTs

ArkadiyStena:
# Summary

Extension for [NFT Standard](https://github.com/ton-blockchain/TEPs/blob/master/text/0062-nft-standard.md).

A standard interface for non-fungible tokens which allow to destroy them similar to [SBTs] (https://github.com/ton-blockchain/TEPs/blob/master/text/0085-sbt-standard.md#3-destroy). Destroyed NFTs looks like they were never minted, and there is no possibility to re-mint them

# Motivation

Sometimes people want to get rid of an NFT in their wallet, but don't want to sell or gift it to anyone (for example, if the creator of NFT collection accidentally minted an extra NFT), and the only thing they can do is to send the NFT to a zero address. However, this solution has two problems:
* Sending an NFT to a zero address is a loss for the user (a gas fee must be paid and the balance of the NFT smart contract cannot be withdrawn).
* Indexers will continue to display this NFT as if it belongs to some user, which means the displayed supply of the NFT collection could be reduced, even though no actions can be performed with this NFT anymore.

### DIFF
--- a/text/0000-nft-destroyable.md
+++ b/text/0000-nft-destroyable.md
@@ -58,11 +58,11 @@ Ignoring this or any similar standart for burning NFTs will affect users who wan
 
 # Prior art
 
-TODO
+-
 
 # Unresolved questions
 
-TODO
+-
 
 # Future possibilities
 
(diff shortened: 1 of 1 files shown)


chiliec (approved):
Looks good, this functionality is definitely needed!

MrG9090 (approved):

MrG9090 (approved):

MrG9090 (approved):

PR #130, opened 2023-12-14
TEP: create Rebase Jetton standard

IgorErkin:
This pull request follows the corresponding bounty task: https://github.com/ton-society/grants-and-bounties/issues/342

### DIFF
--- a/text/0000-rebase-jettons-standard.md
+++ b/text/0000-rebase-jettons-standard.md
@@ -1,0 +1,399 @@
+- **TEP**: [0](https://github.com/ton-blockchain/TEPs/pull/0) *(don't change)*
+- **title**: Rebase Jettons standart
+- **status**: Draft
+- **type**: Contract Interface
+- **authors**: [Igor Erkin](https://github.com/IgorErkin), [Konstantin Komarov](https://github.com/KonstantinK01) 
+- **created**: 04.12.2023
+- **replaces**: -
+- **replaced by**: -
+
+# Summary
+
+A standard interface for Rebase Jettons. 
+
+# Motivation
+For the rebase tokens, balance of a user is calculated as follows:
+
+`balance = shares * total_supply / total_shares;` (Equation 1)
+
+`shares` - user share of total supply, 
+`total_supply` - sum of all tokens, 
+`total_shares` - sum of all user shares.
+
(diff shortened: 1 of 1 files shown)


EmelyanenkoK:
If I got it right, the aim of this TEP is to make it possible to increase "displayed balance" for all holders simultaneously.
While I'm not sure it is indeed needed, can the same purpose be implemented like here: https://github.com/ton-blockchain/TEPs/pull/98
In particular will fractional decimals solve this less "invasive"? In particular, instead of `"decimal":"9"` you specify "fractional_decimal":"734555" and it means that for displayed balance wallet/explorer need to divide number of indivisible units by `734555`. By changing "fractional_decimal" on Jetton Root all balances will be multiplied by a corresponding factor.
Advantage of "change metadata" approach is better compatibility with existing software.

If we are concerned that display on services that understand fractional decimals and not are different, it is possible to further introduce "fraction_decimal_symbol". In that case metadata will looks like:
```
{
   ...,
   "symbol": "RAWMyJett",
   "fraction_decimal_symbol": "MyJett",
   "decimal":"0"
   "fractional_decimal":"734555"
}
```
Those services that understand "fractional_decimal" will show MyJett rebased balance, those that don't will show not-rebase balance of RAWMyJett.

IgorErkin:
@EmelyanenkoK , thank you for your feedback!

Yes, you are correct about the purpose of this TEP. It is needed to implement rebasing mechanics, which imply changes in the token supply, and the corresponding changes in the balances of all holders.

Regarding your proposal, the only concern I have with this approach is that the balance value of the wallet will need to be calculated by each front-end application individually. Given that there are numerous front-end applications developed by different teams, there is a potential for discrepancies in calculation methods, such as using different data formats or rounding rules. In this regard, the use of getters seems preferable, as the calculation takes place in them and they act as SSOT.

So, we can add it, but at the same time I propose to also keep the getters.


A few words about why we need this standard:
1. It is the only way to implement yield-bearing tokens that are 1:1 equal to and backed by the underlying asset.
2. In other networks, there are already dozens of similar projects with billions $ in TVL. 
3. The trend towards RWA tokenization with a potential of trillions of $. This standard will play a significant role in this trend. For example, ondo.finance, one of pioneers in this direction founded by managers from Goldman Sachs, plans to implement rebasing mechanics for USDY (one of their products). Or, the first RWA project to come to TON - Ledgity Yield protocol, also plans to implement rebasing mechanics.
4. At least 5-6 in the public field want/wanted to issue rebasing tokens in the TON blockchain

PR #131, opened 2023-12-26
Referral code based on Query ID

oleganza:
This is a proposal to split query ID in two halves: app identifier and the random app data.

[Specification](https://github.com/tonkeeper/TEPs/blob/queryid/0000-referral-query-id.md)

### DIFF
--- a/0000-referral-query-id.md
+++ b/0000-referral-query-id.md
@@ -16,7 +16,7 @@ Standard way to self-identify apps that initiate transactions makes it easier to
 # Guide
 
 * Applications and services may self-assign unique identifiers to disambiguate their application from others.
-* In privacy-preserving applications users may expect absence of any referral codes. In such cases apps may offer an option to leave the referral code as all-zeroes or use other app’s well-known identifier at random.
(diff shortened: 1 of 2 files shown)


mr-tron:
Please add algorithm for creating first part. I suggest to user  CRC32 IEEE like for tlb schemes. 
for example
```
crc32('tonkeeper') = 0x546de4ef
crc32('dedust') = 0xc671412d
```

EmelyanenkoK:
I think it need to be explicitly mentioned that this data can be used only for offchain analysis and can not be fully trusted.

oleganza:
@mr-tron @EmelyanenkoK added notes as you suggested. Please review.

howardpen9:
looks cool for me.


But what is the impact for this to DeFi like STON.fi and DeDust?

I remember they use query_id in zero only I am afraid.

ProgramCrafter:
> Second 32 bits (low bits) are reserved for arbitrary use: identifying individual queries in smart-contracts...

By the way, is there *any* onchain usecase which would allow to "identify individual queries" securely? I have not seen one so I guess main (and probably the only) usecase for query IDs is optimizing blockchain indexers.

PR #135, opened 2024-01-08
fix brocken link

mbaneshi:
https://github.com/romanovichim/TonFunClessons_Eng/blob/main/arhive/9lesson/ninthlesson.md


EmelyanenkoK:
Thank you for fix!
Seems like #138 is also fixed this issue.

PR #146, opened 2024-03-19
Add TEP404 standard

kojhliang:
TEP404 standard is inspired by [ERC404]((https://github.com/Pandora-Labs-Org/erc404)).   
This protocol can increase the liquidity of NFT transactions, achieve fragmentation of NFTs, and facilitate the exchange between different NFTs.

### DIFF
--- a/text/0000-trc-404-standard.md
+++ b/text/0000-trc-404-standard.md
@@ -1,0 +1,392 @@
+- **TEP**: [404]
+- **title**: Semi-fungible token standard 
+- **status**: Draft
+- **type**: Contract Interface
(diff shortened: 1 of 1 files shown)


kojhliang:
Thanks @howardpen9 @sidebyandrew  contribution

tactfunc:
<3

howardpen9:
good to heard

ProgramCrafter:
Why NFTs are needed in this case? Why not use simply jettons with corresponding render type?

NFT sale (for fixed price) can be replicated on jettons as well.

PR #154, opened 2024-05-20
Adding on-chain libraries TEP

kehiy:
I wrote this proposal to introduce a way to make a blockchain more scaleable and lightweight.
Maybe some part of this proposal need to be changed and it needs more discussion to have final changes and versions.

> Note: I Used empty TEP number to be considered.

### DIFF
--- a/text/0000-on-chain-libraries.md
+++ b/text/0000-on-chain-libraries.md
@@ -10,40 +10,100 @@
 # Summary
 
 This document introduces a new type of on-chain source codes with no state that their functions can be invoked by 
-their address and function identifier from contracts.
+their address and a function identifier from other contracts/actors.
 
 # Motivation
-
-Which problem we address? Why it is important to make a new TEP?
+ 
+This proposal introduces a way to deploy libraries (e.g. TON stdlib written in FunC) one and use them many times on the
+network. this will reduce the size of network, make smart contracts more light weight and managing and using libraries
+for contracts easier.
 
 # Guide
 
-Explain this document in simple language, as if you were teaching it to another developers. Give examples how your feature will work in real life.
+When we write a smart contract we are able to import other source code in our code and use their functions and methods.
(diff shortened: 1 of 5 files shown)


kehiy:
@tolya-yanot @EmelyanenkoK @hacker-volodya 

Hi, can you have a review guys? should I publish this in any specific place?

ProgramCrafter:
I shall note that this is already implemented, only that it's currently inconvenient to call specific functions out of library; thus, the changes would only be within FunC compiler (though, groups of functions are more suited to Tact).

---

> 1. The most used TON library (stdlib.fc) can be deployed on-chain which is used in most of the contracts (including wallets).

It would *increase* fees, as stdlib is already written in TVM assembly (and something is even emitted by FunC compiler).

kehiy:
@ProgramCrafter, thank you for your response. so, you are saying that at the current moment, there is no duplicated stdlib code on TON blockchain? or I misunderstood it?

text/0000-on-chain-libraries.md
ahmadrafiee331, line 1: UQBhtq4l0mRuN92W5Ak1ILjIrIswuLPUpVKSUC4jaST0qgE5

ahmadrafiee331 (approved):
UQBhtq4l0mRuN92W5Ak1ILjIrIswuLPUpVKSUC4jaST0qgE5

EmelyanenkoK:
> @ProgramCrafter, thank you for your response. so, you are saying that at the current moment, there is no duplicated stdlib code on TON blockchain? or I misunderstood it?

Stdlib.fc generally doesn't actually contain new code, it consists of useful funC aliases to asm instructions.

Library cells as reusable and cheap to storage references to constant "chunks" of data (it also may be treated as chunks of code) is already  implemented, you can find it in the docs: https://docs.ton.org/develop/data-formats/library-cells

PR #371, opened 2024-11-24
Create non-disruptive_fee_updates.md

EmelyanenkoK:


AfeworkShalom09 (approved):

AfeworkShalom09:
- I can't to view all my projects

AfeworkShalom09 (approved):

Mortezanouri1378:
https://github.com/ton-blockchain/TEPs/blob/master/text/0074-jettons-standard.md

EmelyanenkoK:
From discussion with SpyCheese: Limited depth of chains for "fixed fees" is bad, because malicious actor may prepare chain in such a way that 1024 depth will be reached in the middle of 3rd-party protocol (DEX for instance) causing disruption of this protocol.
We may neglect issue of infinite chains with old configs and keep copy of gas-related configs in metadata_envelope: it will be deduplicated anyway and solves the problem of storing historical config param settings

soheilbarakt:
scam why you cancel pr at me

PR #467, opened 2025-04-06
Create 0467-normalized-message-hash.md

EmelyanenkoK:


mr-tron:
LGTM

NikitaZimens (commented):
я не а

TEU7ZoTHLa7sZchXmdqZn5pE75ikooVNfv (commented):
alvandisaeed43@gmail.com

EmelyanenkoK:
@a2468834 Issue with trace-id (which is essentially the id of first transaction in trace) is that before message reached the blockchain we do not know the trace-id, for now, giving latencies of delivery and indexing it may require up to 20 seconds that is detrimental for UX. It is also not convenient for services : for instance some service made withdrawal and want to both insert it into database and give user link to explorer.

Since explorers (tonscan and tonapi) now can show pending transactions immediately (with latency as low as 200ms) after external reached explorer (not commited to blockchain, but merely got to indexers) it is expedient to provide the way how user can get identifier for such pending events.

a2468834:
Okay, I see. Waiting for the first transaction hash might be a pain point for some of services.

However, I'm not quite getting the point about why do you choose the current definition of norm message hash? I think it will not be significantly slower if we calculate cell hash of the "full" content of external message. Besides, both choices have the same bit length (i.e., 256 bits), so using hash of full external message would consume the same bits as the norm message hash.

ardh101532:
> 

ARD

text/0467-normalized-message-hash.md
ardh101532, line 1: AR

ardh101532: > AR

Dragos130298: daaa

Dragos130298 (commented):
daaa

PR #496, opened 2025-06-20
add "Configurable Bounce Message Size" proposal

NickNekilov:
This TEP introduces a mechanism for senders to specify how much data should be included in bounce messages when transactions fail, allowing for more flexible error handling while maintaining backward compatibility with existing contracts.


BeyCoder:
We have the same issue with user-friendly errors in XDAO.app. Our complex design flow makes it more challenging to process errors with a 256-bit limit, as we can't include a single address in bounce messages.

However, this proposal will help us resolve the issue and reduce the time spent on developing error processing.

PR #524, opened 2025-09-18
Add Tokenized Vaults Standard TEP

ipromise2324:
This TEP establishes a standardized interface for tokenized vaults on TON, enabling consistent deposit/withdrawal operations, rate queries, and cross-protocol communication. Inspired by ERC-4626 but adapted to TON's asynchronous design, it improves DeFi ecosystem interoperability and reduces integration complexity.

PR #525, opened 2025-09-19
Tokenized Vaults

krigga:
This spec and reference implementation were co-designed with Tier-1 partners (Aave, Ethena, Affluent) on real business cases. We’ve already collected their comments and support, and implemented the majority of that feedback in the current draft.

We’d appreciate your review and comments.


ipromise2324:
I've posted a comprehensive technical analysis comparing TEP-524 and TEP-525 in the #524 discussion thread. The comparison examines both proposals across 9 key technical dimensions.

Please review the analysis here: https://github.com/ton-blockchain/TEPs/pull/524#issuecomment-3313728256

Looking forward to community feedback on the technical merits of both approaches.

PR #529, opened 2025-09-24
fix: changed outdated link for TEP 74 Jetton contract

fruitful1l:
The link used to go to the first and outdated Jetton implementation.


zahrasohkandan-pixel:
همه ادغام شود

PR #538, opened 2025-10-22
fix: rename `forward_amount` to `forward_ton_amount`

verytactical:
Jetton TEP mentions `forward_amount`, but TL-B defines only `forward_ton_amount`.


spinnercasino12345-ux (commented):
suggested, typed out and provided an alternative source for option

frankrking36:
Can we move full acount over to present address

PR #539, opened 2025-10-22
Add TEP-539: Fundamental Domains

ProgramCrafter:
This document introduces Fundamental Domains in the .ton zone, with the first one - `explor.er.ton` - enabling dApps to access blockchain information via a proxy instead of centralized APIs.


Rastakhcarbon:
🤨

Issue #418, opened 2024-12-27
License?


Maniues:
This repository is unlicensed. Maybe there should be a license for TEPs (as they are standards for TON).
My suggestion is CC-BY, MIT, CC0 or another Free License.

I can make a PR if you agree to one license.

Sergehima333:
Ничего не понял ,но ладно

Pikasso80:
> This repository is unlicensed. Maybe there should be a license for TEPs (as they are standards for TON). My suggestion is CC-BY, MIT, CC0 or another Free License.
> 
> I can make a PR if you agree to one license.

Navvabk:
0x17afe

coexglobal:
Очередной разводняк без лицензии без подтверджений и тп

rajtuhin317:
Nice

ashyking:
ok

Cheeezzyy:
Sel, 14 Jan 2025 pukul 21.17 ashyking ***@***.***> menulis:

Sergehima333:
Hoo are you?

пт, 17 янв. 2025 г., 05:21 yalan2ny ***@***.***>:

Cheeezzyy:
> This repository is unlicensed. Maybe there should be a license for TEPs (as they are standards for TON). My suggestion is CC-BY, MIT, CC0 or another Free License.
> 
> I can make a PR if you agree to one license.

Tooniyam:
> Hoo are you?https://github.com/ton-blockchain/TEPs/issues/418/
> 
> пт, 17 янв. 2025 г., 05:21 yalan2ny ***@***.***>:
> […](#)

Elgayar777:
LFG

Koki1610:
.

bhie04:
21

bhie04:
21

karpovamarina618-blip:
### 

> _[](url)_

PR #556, opened 2025-12-12
Add license for template

Maniues:
According to #418 this repo has no license.

Because relicensing all existing TEPs requires permission by the authors of all these TEPs, I started by adding it to the template, so next TEPs will be licensed that way.

Why license?
Many Open Source projects (including blockchain projects) add license for their "TEPs". For example:
- Ethereum: CC0 - https://github.com/ethereum/EIPs/blob/master/LICENSE.md
- Bitcoin: BSD, Public Domain (some BIPs are unlicensed)
- BSC: CC0
Even programming languages, e.g. Python (PEP) are placed in the public domain.

Why CC0?

It is a popular option in strong ecosystems like Ethereum, BSC, and it is perfect for that proposals.

For this change, I need only @hacker-volodya approval, as the only author of template.

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -46,3 +46,7 @@ If there are some questions that have to be discussed during review process or t
 # Future possibilities
 
 Do you have ideas, which things can be implemented on top of this TEP later? Write possible ideas of new TEPs, which are related to this TEP.
+
+# Copyright
+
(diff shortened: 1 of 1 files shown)


Ascalad:
A pull request is created to fix wallet receipt issues caused by not having a repository license.

